PR #295, opened 2014-02-28
Tables with <thead> and <tbody> tags

rindeal:
table header is now between `<thead>` tags and table body is between `<tbody>`

this allows great javascript and css magic with tables, especially sorting


rindeal:
Closing in favor of #410

PR #320, opened 2014-11-12
Guest network support for Gargoyle

yarrouye:
Eric,

I would like to contribute support for guest networking in Gargoyle, something that has been asked for by many. See the [applicable thread](http://www.gargoyle-router.com/phpbb/viewtopic.php?f=14&t=6358#p25768) on the Gargoyle forums. The UI mirrors that of the AP network for SSID, encryption, SSID broadcasting and client isolation. By default, clients are isolated. Also, it is possible to only set a guest network on one of the frequencies when the AP is in dual mode.

As I posted there, it seems (from posts from pbix) that the driver you use for Broadcom chipsets won't support that, so for now non-Atheros chipsets will not show the guest network display. This is controlled by the variable `gns` (guest network supported) in `setWifiVisibility()` in the `basic.js` file.

Let me know if you have any questions.

screen shot 2014-11-10 at 8 12 28 pm
screen shot 2014-11-10 at 8 13 02 pm
screen shot 2014-11-10 at 8 13 26 pm
screen shot 2014-11-10 at 8 13 42 pm


ericpaulbishop:
I'm working on merging this now (finally), and obsy's suggestion is a really good one for reasons other than even what he is stating above. 

To clarify his reasoning (if I'm understanding it properly), he is saying that since a guest network is often open, Google will scoop up this data and use it to help them track you if you're connected to this network, which is bad and often not wanted.

However, there's an even better reason, and one which suggests the MAC should be randomized automatically: Your current code for detecting which interface is a guest network is really kind of a hack. In particular, it looks like it will fail if you have an ap+client configuration. Searching on the explicitly set MAC of a guest network is a much cleaner approach than just taking all wlanX.1 interface to be guest networks.

yarrouye:
> However, there's an even better reason, and one which suggests the MAC
should be randomized automatically: Your current code for detecting which
interface is a guest network is really kind of a hack. In particular, it
looks like it will fail if you have an ap+client configuration. Searching
on the explicitly set MAC of a guest network is a much cleaner approach
than just taking all wlanX.1 interface to be guest networks.

Sounds good to me! I agree my current method is a bit of a hack and if we
know the MACs of the guest network(s) (there could be one per frequency),
or had any other way to mark which ones are guests, it would be much more
robust.

PR #326, opened 2015-01-03
usb-storage: fstab

obsy:


ericpaulbishop:
What's wrong with the current blkid command? blkid is installed and works fine.

obsy:
Nothing special. I just think that we should use the tools from fstools

ericpaulbishop:
I disagree. 

A problem I've seen before is that when a version of a tool shifts, the output can shift slightly. That  can result in parsing errors any time the version of the tool gets bumped. For something simple like describing a disk, which really isn't a task that should change at all even as we go to newer versions, I'd rather use a tool which I can easily keep fixed to a specific version, with consistent output. I don't really want to keep ALL of fstools locked to a specific version, just the disk description output we are parsing here. Using the blkid tool to generate this output, which has its source the repo, locally, and therefore won't change, seems reasonable.

PR #336, opened 2015-02-12
translation update

tutsde:
German translation updated


tutsde:
German translation updated

ericpaulbishop:
I've merged your patch-1 branch with the basic.js updates, but your openvpn.js updates do not merge cleanly.

There are already translations for some of the variables you added. I don't speak German so I don't know if your translations are better. If you want to send me updates to openvpn.js, though, can you merge with my master branch, re-do the translations (or pick the version that is translated best) and send me a pull request again? That should allow me to merge cleanly without having to decide which translation is best myself, a task I am definitely not qualified to do.

PR #341, opened 2015-04-20
Refresh minidlna database...

lantis1008:
on reboot. It seems to have a problem with the lan interface being down, this triggers a refresh when it is ready. 

I'm unsure if i have correctly modified the makefile also, it's my first attempt at such an implementation.

This has only been tested on my system at home (WNDR3800)


lantis1008:
causes instability. disregard. will keep searching.

PR #343, opened 2015-04-27
fix empty minidlna on reboot

lantis1008:
can you double check i've modified the makefile correctly please eric? i'm new to makefiles.


ericpaulbishop:
The Makefile looks fine... but it doesn't look like you're properly testing for minidlna being enabled.

You should test this way:

minidlna_enabled=$( uci get minidlna.config.enabled 2>/dev/null)
if [ "$minidlna_enabled" = "1" ] ; then
....
fi

lantis1008:
it's a supported switch case for the init.d implementation in openwrt according to the wiki, but i agree that using uci is a more robust check. i'll make that change.

So what is the best way from here? close this one, edit, new pull request?
or can i edit this one as is?

ericpaulbishop:
If you push a new commit to the same branch, I believe the pull request
will automatically be updated.

ericpaulbishop:
Also, while you're right that the enabled init script functionis implemented, the problem is that in the case of minidlna, the init script, even if it is enabled, checks the uci variable and simply exits if it is set to 0. Therefore you need to check the uci variable in order to determine if minidlna is actually enabled/disabled.

lantis1008:
ah yep that makes sense. applied your suggested changes. thanks for your assistance.
i much prefer working on this stuff than my university projects lol.

PR #353, opened 2015-06-19
Update miniupnpd.init

lantis1008:
From my testing, the source of Phgerin's problem (http://www.gargoyle-router.com/phpbb/viewtopic.php?f=14&t=7227#p29945) is that after a sysupgrade (preserve settings) the firewall gets a little upset.
A simple restart of the firewall fixed these issues.
Steps to fix:
1. disable upnpd
2. issue firewall restart
3. reenable upnpd

so two proposed solutions, this fix here. 
OR
making sure that if we do a "preserve settings" the firewall gets properly flushed and restarted.

What do you think?
I have yet to test this change as i am still doing the 4GB transfer test.

### DIFF
--- a/package/miniupnpd/files/miniupnpd.init
+++ b/package/miniupnpd/files/miniupnpd.init
@@ -205,4 +205,5 @@ stop() {
         [ -x /usr/sbin/ip6tables ] && {
 	    ip6tables -t filter -F MINIUPNPD 2>/dev/null
         }
+        /usr/lib/gargoyle/firewall_restart.sh
 }


ericpaulbishop:
I don't think restarting the firewall in the miniupnpd init script is the answer -- that sounds like it might introduce the danger of an infinite loop where each script keeps triggering the other.

I haven't seen miniupnpd not getting restarted properly after a normal reboot. I am rather curious why it might not be coming up properly after a sysupgrade -- nothing in the config file has changed, just the binary.

lantis1008:
I've noticed a few odd behaviours on sysypgrade but never thought much of them. Maybe 1/3 upgrades, the graphs for bw distribution and qos distribution do not function. Firewall needs to be restarted to bring it to life. Could be a related symptom.

You can close this unless you wish to keep discussion open here?

PR #362, opened 2015-08-27
add mvebu to build, update switchinfo, update dir835 patch...

lantis1008:
add mvebu to build, update switchinfo, update dir835 patch, add new dns provider, fix Portuguese language problem, bring adblock changes to CC.

these are my stable changes.


lantis1008:
I reverted my changes, but kept some of the additional code to suit mvebu. Apologies.

lantis1008:
https://forum.openwrt.org/viewtopic.php?pid=289535#p289535

See here. It's being brought in line with the ar71xx implementation. 
I'll leave the code for now. And it can be removed at a later date if we desire.

PR #366, opened 2015-09-09
Fix Adblock plugin conflict with DDNS

lantis1008:
Adblock remote list includes tracemyip.org as an "ad" and therefore blocks it. This results in the remote address being updated to 127.0.0.1 when the url loop reaches that entry.
Whoops!


nworbnhoj:
should you not whitelist all of the URL's 
in default_ip_lookup_url_data 
in check_local_ip_urls.c ?

lantis1008:
Just trying to keep size to a minimum. Only add what we need. I now know the symptoms that occur when this problem happens so i'll know what  to fix next time. If the addresses weren't embedded in the ddns application then i could do something fancier with them but i think this solution is suitable.

PR #385, opened 2015-10-26
Move to html5

stopspazzing:
This is the only change needed to move to html5. I have tested and confirmed on all pages without issue: https://i.imgur.com/CqYm7Jt.png
TP-LINK WDR3600

No other changes are needed for html5 support. It's built into modern browsers. http://www.w3schools.com/html/html5_browsers.asp
I haven't tested screen readers, but I don't see why it would be an issue for them.

Line 321 may need language support, but doesn't show up on user side; it's meant for search engines to use as the description below the links, can be omitted.

The 2 other lines in the head are for responsive design layouts. Can also be omitted, but prepares code for responsive support.

### DIFF
--- a/package/gargoyle/src/gargoyle_header_footer.c
+++ b/package/gargoyle/src/gargoyle_header_footer.c
@@ -314,12 +314,14 @@ int main(int argc, char **argv)
 		
 
 
-		printf("<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.1//EN\" \"http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd\">\n"
-		       "<html xmlns=\"http://www.w3.org/1999/xhtml\">\n"
+		printf("<!DOCTYPE html>\n"
 		       "<head>\n"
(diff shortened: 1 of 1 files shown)


stopspazzing:
You could move away from awful insecure cookies and use sessionStorage without the need for hashing. This will shrink the code not expand it. This would also open up other features like the input attributes (required,regex checks,min and max characters allowed, etc), but with those support is IE9+ so would still need current code as fall back. You could cache all static variables to sessionStore on login, then read from it instead of pulling from the device. Then if any page gets saved, update those variables in sessionStore after pushing them to device. 

A simple:
```javascript
sessionStore.LoggedIn = "1";
sessionStore.TimeStamp = timestamp;
```
on login and on log out:
```javascript
sessionStore.LoggedIn = "0";
sessionStore.TimeStamp = "0";
```
then check it on pages:
```javascript
var check = sessionStore.TimeStamp + //15mins...;
if(sessionStore.LoggedIn == "1" && currenttime < check){
//continue with displaying page...
}else{
//redirect to login page...
}
```
None of those cookies. This is saved per protocol AND per tab. Not accessible by anything but originating domain and protocol, aka 100% secure.

PR #389, opened 2015-11-10
Tweek Plugins page

nworbnhoj:
Some users find the "Refresh Plugins" button hard to locate at the bottom of the page. Move it to the top of the section to improve visibility.

Some users have not realized that a WAN connection is required to Refresh Plugins. Show a red prompt if the user clicks the Refresh Plugins button without an active WAN connection.


nworbnhoj:
Tested this again just now and there is a problem that I missed the first time.  Will have another look tomorrow

PR #390, opened 2015-11-10
Tweek Plugins page

nworbnhoj:
Some people have had trouble locating the "Refresh Plugins" button at the bottom of plugins.sh. Move the button to the top of the section.

Some people have not recognised the need to have an Internet connection in order to Refresh Plugins.  Show a red prompt to "check Internet Connection" when the "Refresh Plugins" button is clicked when there is not a currentWanIp address.

[EDIT] Added i18n stuff

### DIFF
--- a/package/gargoyle/files/www/js/plugins.js
+++ b/package/gargoyle/files/www/js/plugins.js
@@ -411,8 +411,8 @@ function updatePackagesList()
 		document.getElementById("wan-warn").style.display = "inline";
 	} else {	
 		document.getElementById("wan-warn").style.display = "none";
- 		var e=["opkg update"];
-		execute(e)
+ 		var cmd=["opkg update"];
+		execute(cmd)
  	}
 }
 
--- a/package/gargoyle/files/www/js/plugins.js
+++ b/package/gargoyle/files/www/js/plugins.js
@@ -405,8 +405,15 @@ function uninstallPackage()
 
 function updatePackagesList()
 {
-	var cmd = [ "opkg update" ];
-	execute(cmd);
+ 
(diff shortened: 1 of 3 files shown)


nworbnhoj:
For the sake of simplicity, the check for Internet connectivity is done with a check for WAN IP which is simple, but not definitive. ie the red prompt may show when there is Internet access via a non-WAN port

nworbnhoj:
If the router has access to the Internet via a LAN port, then the red prompt WILL show AND the plugins WILL refresh. However, this prompt is for users (see forum) who need the prompt!

An alternative would be to change the prompt wording to "There is no Internet connection on the WAN port". We could also make the connectivity check more sophisticated, (by say pinging Google, or checking the opkg response for "WARNING") but this adds complexity and bulk that is marginal.

krzyko:
I've missed "separate WAN IP check from opkg update" patch. With it plugins will update also with WAN  port disabled.

lantis1008:
I think this needs additional work.
Due to the nature of the "execute(cmd)" function in common.js, at the end of the updatepackages function, the web-page refreshes. This means that the WAN warning only remains visible for < 1 second.

Several options:
1. Move the check for the WAN into the resetData() function so that it is checked on page load every time.
2. Make it a javascript alert style message instead of just a text on a page.

Option 2 is more invasive than 1 and less preferred due to the fact that this check can fail even though a successful package update has occurred.

nworbnhoj:
Strange that I did not notice "This means that the WAN warning only remains visible for < 1 second."
Agreed - I have made the [change](https://github.com/nworbnhoj/gargoyle/commit/01d1ef0ec3c908dd35f86ce91d47e8aefe1c2c65)

PR #396, opened 2015-11-26
Update gargoyle interface to support 802.11ac

lantis1008:
Along with this i ended up majorly rewriting a lot of code so please scrutinize this pull request carefully.

If you have any feedback for what should change i would be happy to make the appropriate changes, i have a lot of free time currently.

In particular i have not been able to test devices with only single band wireless.


lantis1008:
going to work on not using a persistent file. would be better space wise.

ericpaulbishop:
Sorry for the delay in looking at this -- thank you for your work here, this is great. I need to spend some time testing this fully, but in general, it looks good.

lantis1008:
Bugs found. Whoops! Working on them.
Found a way of testing single band devices which i hadn't thought of.

lantis1008:
routers that only have a single radio (2.4ghz) should behave properly now. God help us if we ever support a router that only has a 5ghz radio.

lantis1008:
along with the latest commit, 06f83f1a88cab9aa2b209c6cb8a4caefad022278, scan_wifi.sh needs updating, however i'm struggling to come up with something that works properly

lantis1008:
How is testing coming along with this?
More and more people are showing up with AC devices on the forums and some of them are a bit beyond the tinkering and tweaking required to do this manually. 
If I can assist at all please let me know.

ericpaulbishop:
It will be done by the end of the week.

lantis1008:
Cool. I'll avoid making further commits to this Pull Request then. 

My current bug list (in case it doesn't match yours):
- scan_wifi still uses iwconfig
- bridge mode still has some display issues, mostly in non ht_capab devices.

ericpaulbishop:
Just FYI, "end of the week" may have been a bit ambiguous. I was originally trying to finish today, but it may be Saturday or Sunday before I merge the wireless_ac branch (which is now a separate branch in my repo, should be there on github), into master

lantis1008:
No worries. 

I'm just glad the base code could be of some help.

ericpaulbishop:
This is definitely VERY helpful -- thank you.

PR #397, opened 2015-11-27
Use "iwinfo" instead of "iwconfig"

lantis1008:
on smaller devices we removed wireless tools which took away "iwconfig" and broke the hosts page on the GUI.
we still have "iw" and "iwinfo" available to us.
similar changes will need to be made for the madwifi case however i do not have any devices to test with.


obsy:
@lantis1008: could you also change plugin-gargoyle-wifi-schedule?

lantis1008:
Yep I'll take a look.

PR #398, opened 2015-11-27
Bugfix and Simplify UCIContainer

nworbnhoj:
The code has been simplified by:
* replacing the use of UCIContainer.listOptions with checks to see if (UCIContainer.value
instanceof Array). 
* replacing the use of UCIContainer.keys with checks on values.hasOwnProperty(key)
* bugs identified and fixed.

Background. While implementing new functionality to store uci list
values, problems were encountered. While UCIContainer.set() accomodated
for uci list values using UCIContainer.listOptions,
UCIContainer.getScriptCommands() did not refer to
UCIContainer.listOptions and did not return the correct commands to set
uci list values. While investigating and correcting this problem,
various other corrections and simplifications were undertaken.

UCIContainer.values is declared as an Object rather than an Array
UCIContainer.listOptions is replaced with instanceof Array checks
UCIContainer.keys is replaced with  values.hasOwnProperty(key) checks
UCIContainer.set() is recoded
UCIContainer.clone() corrected clone of Array values

### DIFF
--- a/package/gargoyle/files/www/js/common.js
+++ b/package/gargoyle/files/www/js/common.js
@@ -1,8 +1,8 @@
 /*
- * This program is copyright © 2008-2011 Eric Bishop and is distributed under the terms of the GNU GPL 
- * version 2.0 with a special clarification/exception that permits adapting the program to 
+ * This program is copyright © 2008-2011 Eric Bishop and is distributed under the terms of the GNU GPL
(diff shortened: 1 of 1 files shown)


nworbnhoj:
The diff is very difficult to interpret :-(  
Any thoughts how to tidy this up?
I have core.autocrlf=input as recommended on Linux.

PR #399, opened 2015-11-29
Cleaning conditional directives that break statements.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/package/httpd-gargoyle/src/httpd_gargoyle.c
+++ b/package/httpd-gargoyle/src/httpd_gargoyle.c
@@ -7,7 +7,7 @@
  * 		  	The original mini_httpd was created by Jef Poscanzer
  * 		  	http://www.acme.com/software/mini_httpd/
  *
- *  Copyright © 2008-2011 by Eric Bishop <eric@gargoyle-router.com>
+ *  Copyright ï¿½ 2008-2011 by Eric Bishop <eric@gargoyle-router.com>
  * 
  *  This file is free software: you may copy, redistribute and/or modify it
  *  under the terms of the GNU General Public License as published by the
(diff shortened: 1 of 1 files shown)


ericpaulbishop:
Gargoyle is no longer using the httpd_gargoyle daemon, it's been deprecated in favor of uhttpd. I admit it's still in the repo, but it's not used/maintained anymore.

RomeroMalaquias:
Hello @ericpaulbishop   i'm from the same research team of flaviommedeiros and i need to ask if you prefer this code before or after the commit. What do you think about this kind of change? 
Thanks in advance.

RomeroMalaquias:
I mean, would you accept if it was not for the fact we submitted the pull request to a deprecated code?

ericpaulbishop:
It would depend on what exactly was in the pull request. I review pull
requests on a case-by-case basis.

This one in particular was rejected because it was a deprecated piece of
code.

Eric

RomeroMalaquias:
@ericpaulbishop ,
I see
So to clarify, do you agree with the changes?
Sorry for insisting, I'm doing a search and your response is very valuable.

ericpaulbishop:
I did not take the time to investigate the changes, because it would have
been a waste of time, as this component is outdated.

I do not intend to take the time now.

I therefore cannot give you an answer.

Eric

PR #402, opened 2015-12-15
plugin-gargoyle-email-notifications: Bug and cosmetic fixes

dpint:
Fix for a bug that was pointed out by forum user charlie regarding file
permissions.


lantis1008:
Another option instead of chmod is adding an "sh" to the beginning of all calls to the file. 

i.e. sh /usr/lib/gargoyle/email.sh

Nothing wrong with what you did though

dpint:
Yeah, I know, but it would also require some changes to javascript.

PR #403, opened 2015-12-15
Known devices

nworbnhoj:
The idea is to be able to set policy (Quota, QoS, Restrictions etc) over a Group of Known Devices (owned by an individual for example). To achieve this, the Known Devices need to be identified by their MAC address' and then arranged into Groups. Both Devices and Groups can be named by the Gargoyle Administrator.

Up until now, the best way for a Gargoyle admin to apply policy to users with a number of devices has been to assign a static IP address to each device such that each users' devices are in an IP range. Then, the policy is applied to the IP range, and the IP range is displayed throughout the Gargoyle GUI. Known Devices and Groups enable the Gargoyle Admin to work with meaningful names rather than IP ranges, and relieves the need to assign static IP addresses.

Status
- Gargoyle-Connection-Devices has a Section for Known Devices and another for Device Groups.
- Gargoyle-Firewall-Quotas-AppliesTo-OnlyTheFollowingHosts will accept a Device Group.

Technical
- Approx 17k in size
- Known Devices are stored as a host in uci /etc/config/dhcp
- Each host has a suplimentary Group section
- Each Device Group is represented by an ipset (each set = 550 bytes min)
- IP addresses in each ipset are dynamically adjusted when dnsmasq issues a dhcp lease
- Existing iptables quota rules can check against the membership of an ipset


lantis1008:
Look forward to testing this!

ericpaulbishop:
This is one of two really large pull requests that are now pending, that I need to spend some time testing.

I really like the idea, but just a warning that I want to merge and fully test lantis' Wireless AC pull request first, before I will merge this, so it may be a few weeks before I get to this.

nworbnhoj:
OK great - thanks for the "heads up"

There are a number of logical "next steps" to this development. For example, a Device Group:
- should be a valid option in a QoS Classification Rule (upload and download)
- should be available in Access Restrictions (blacklist and whitelist)
- and conceivably, could be used to deliver targeted DNS services to Device Groups

In addition, Device Groups:
- probably make the option of adding individual IP addresses (to quotas for example) redundant. These comma separated lists of IP addresses, add size and complexity to the existing code, which could be removed (restore_quotas.c lines 488-520 for example)
- probably the Static IP table can/should be merged with the new Known Devices table, to again reduce size and complexity

Question: Should I continue to progress these items and add them to the Pull Request, or tally-ho until this first chunk of code changes is (hopefully) pulled?

ericpaulbishop:
For now, keep working. When I finally pull this (likely into a separate branch to start with, after which I'll merge it back to master), you may want to stop for a bit while I test and fix any issues I find, since after that point all pull requests will have to merge cleanly with any changes I add.

nworbnhoj:
Note that, I have departed from the Gargoyle norm for the behaviour of the Edit Button on the Gargoyle - Connections - Devices page. The edit behaviour is hopefully intuitive for the user, but with reduced code size and complexity.

The Google norm Edit button pops up an independent window.

The Devices and Group Edit buttons remove the row from the table, re-populate the original input fields, and re-locates the browser to the top (useful when the table is long). The user may edit the settings and then click the Add button to place the new info back into the table. Nothing is actually saved until the SaveChanges button at the bottom of the page is clicked (so nothing is lost if the user abandons the changes early)

I think this simplified approach works OK for such a simple form??

nworbnhoj:
So I have now completed the basic implementation that I had in mind for Device Groups including Quotas, Restrictions and QoS. I suspect that people will want to use Device Groups in some of the reporting areas like BW Distribution, but that can wait until later.

The idea of using Device Groups to deliver targeted DNS seems doable but the OpenWRT uci documentation of this seems a little weird. I have asked for clarification on the OpenWRT forum (https://forum.openwrt.org/viewtopic.php?id=61613) so hopefully I get a response. This should be doable by going around uci and direct to dnsmasq config, but it would not tie into Device Groups quite so neatly.

I still think there is an opportunity to claw back the size by removing the option of adding individual IP addresses (to quotas for example), but I will wait until people have had a chance to work with Device Groups a bit before I progress this.

nworbnhoj:
Just realized that I have not implemented Groups into the edit dialogs for each of Quotas, Restrictions and 2 x Qos. I will complete this over the e=next day or so and add to this pull-request.

Also (above) it seems like we have some conflicts now so I will endeavour to resolve these.

nworbnhoj:
Finished implementation of Groups into the Rule Edit popups.

Conflicts resolved (whitespace only)

nworbnhoj:
The StaticIP form has now been replaced by the Known Devices form. A Known Device is identified by its MAC address and given a host name and optionally a static IP address. Apart from simplifying the GUI, this has also delivered some space savings. This entire pull-request now has a net size increase of around 10k (after JS compression)

nworbnhoj:
dnsmasq draws Host info from 3 places: uci &  /etc/hosts & /etc/ethers.

The prior commit makes a shift to uci being the primary place to store and manipulate Host information. This shift to uci is important because the association between Host and Group and networkid are all stored within the uci system.

I will now revisit the other few places where Gargoyle draws info from /etc/hosts and/or /etc/ethers and make a similar adjustment.

lantis1008:
This is a good change.

ericpaulbishop:
By the way, I haven't forgotten about this.

My general policy is to merge simpler pull requests first, so that's why other simpler things are getting merged before this. I still think this is a good work and will definitely get around to merging this in the near future.

nworbnhoj:
Thanks for the feedback. I have been running this development for a couple of months now without an issue. 
Also a couple of forum users are toying with it. There have been a couple of points of confusion, but no bugs so far as I can tell.
https://www.gargoyle-router.com/phpbb/viewtopic.php?f=7&t=8318
I appreciate the time and effort required to evaluate such a pervasive pull request, but is there is anything that I can do to assist ....

ericpaulbishop:
For future reference, the fields used to add a new element to a table should NEVER be used for editing. An editor should pop up a new window (now, if you ever want to implement this with javascript rather than popping up a new browser window that would be fine, and probably better -- the old mechanism is a bit clunky I admit). 

For now I'm going to remove the "edit" option for the devices/groups -- it can be added back later so long as this is not done with the initial controls used to add the device to the table.

nworbnhoj:
> For future reference, the fields used to add a new element to a table should NEVER be used for editing.

So is this a stylistic thing, or data integrity, or security or what? Just interested to understand your reasoning?

ericpaulbishop:
First, you're removing the row from the table. That's NOT the behavior expected with edit. If I hit edit on multiple rows in rapid succession the way you have it, I lose data. That's bad -- really, really, really, egregiously bad.

Second, consistency is important. Every other table in gargoyle pops up a window when you hit edit, so that is what should happen here. As I said, if you want to implement in a more modern way without actually creating a new browser window but creating the new element with javascript, that would be fine and probably an improvement. But, wherever there is an edit button, when you click it, a new window should open.

nworbnhoj:
@ericpaulbishop please see bugfix in prior commit which should be included in you Known Devices evaluation branch

fatso83:
Anything preventing this from being merged?

lantis1008:
It probably needs to be rebased for the current code set. 
I never ended up testing it and I don't recall where it ended up as far as Eric's feedback.

nworbnhoj:
This definitely needs to be rebased to fit the 10mths of development since last commit. I have not been following the changes to the UI in detail so I have no idea how much effort might be involved. I have been running the Known Groups branch for these 10+ months with zero issues.

fatso83:
As this is a clean merge without conflicts, according to GH, a rebase should go smoothly AFAIK.

lantis1008:
common.js, all of the .sh files, all of the template files, all of the .css files, at a MINIMUM, will need to be updated to suit the current development.
Please see the other active branches.

lantis1008:
I'm also really against people using programs to help them develop, that try to "auto clean" files. 
It makes the diff's really messy and hard to read what has actually changed.
That would be _my_ opinion though.

krc4267:
Is anyone still working on this? I'd love to have it for my system.

lantis1008:
Not at this stage.
Its now severely out of date.

PR #408, opened 2015-12-20
Quota Usage Improvements

nworbnhoj:
The Quotas Usage page can now display Bytes Used, Percentage Used and
Limit. [Tareq](http://www.gargoyle-router.com/phpbb/viewtopic.php?f=13&t=8048&p=34207&sid=31ccd14d9ed7ae15c9ef947b63f61aed#p34207)


nworbnhoj:
IP addresses limited by an ALL_OTHERS_INDIVIDUAL quota in a prior period
have been hidden to reduce clutter. Similar to [vortex66](http://www.gargoyle-router.com/phpbb/viewtopic.php?f=5&t=8078#p34252)

user501254:
@nworbnhoj, this is nice. I have a suggestion; add another field to indicate if "Shut Down All Internet Access" or "Throttle Bandwidth" has been activated for a particular host when they exceed their quota.

nworbnhoj:
With the 100% now in red it is much easier to see who is at the limit. Maybe a couple of special characters would be a good way to indicate this in the interests of saving space (screen and flash)
⊠ 100%  and ⊡ 105%. Alternatively I could use some other formatting like strike-thru or background color. Thoughts?

user501254:
Characters ⊠ ⊡ may not play well with mobile browsers. 

IMO, having color codes would be better. Like:
- red for those who have exceed quota and no longer have internet access
- yellow for those who have exceed quota but have internet access with throttled bandwidth
- green or black for those who are within their quota limits

Color HEX codes can be similar to those on `hosts.sh` page of the router.

nworbnhoj:
You are right - ⊠ ⊡ no go on mobile browsers. 

Take a look at one of the commits above - the hosts.sh colors are already incorporated with grading from green 0% thru Red 100%.  
Perhaps simple strike-thru on blocked red 100%  and the throttled ones can just continue red beyond 100%

user501254:
Okay, just saw [diff-634fdd5R257](https://github.com/ericpaulbishop/gargoyle/pull/408/files#diff-634fdd5eed6b9f0b22a92a9cd4ae2ea3R257) which incorporates those colors.
I guess strike-through will be nice then.

nworbnhoj:
Turns out .... the info required to indicate blocked/throttled is not readily available in quota-usage.js so it requires a bit more effort than expected. I will put it on my list for a bit later.

PR #410, opened 2015-12-21
Table sorting

nworbnhoj:
I have begun to implement the much requested Table Sorting in Gargoyle based on existing Javascript (GPL3.0) of only 10k size. 

Two tables have been implemented to demo the capability. 
- wol_table is a simple case static table with zebra striping
- hosts_sh is a slightly more involved case with 3 dynamically updates tables with zebra striping

Both proceeded easily with only a very few lines of code.

The next test case will be dhcp_sh hosts_table containing edit and remove buttons. Once this is complete, extending to all other tables in Gargoyle should be a trivial exercise.

Once the hooks are in each table, the 10k gs_sortable.js could easily be delivered as a plugin if need be.


nworbnhoj:
Sorting is now possible on 8 Gargoyle tables.
A few minor changes and extensions were made to gs_sortable.js
The compressed gs_sortable.js < 8k

rindeal:
+1, much better then my #295

nworbnhoj:
Actually it was inspired in-part by [#295](https://github.com/ericpaulbishop/gargoyle/pull/295) and the 2nd commit is almost identical

PR #414, opened 2015-12-28
Fix wifi_schedule plugin to use iwinfo instead of iwconfig

lantis1008:
as requested by @obsy


obsy:
And makefile - drop dependency to wireless-tools

lantis1008:
fixed dependency in #417

Issue #418, opened 2015-12-28
Wireless Client Bridge not working


lantis1008:
Hi Eric,

Lots of reports of wireless bridge not working coming from the forum. Common theme seems to be all reports coming from users with 4MB routers.

Looking at the following patch,
https://github.com/ericpaulbishop/gargoyle/blob/5dd9058f90155b28bbb9a9d838f90586e7cddfe9/patches-generic/011-arpnat.patch

it relies on "iwconfig" as well.
I'm not too confident on the operation of the whole wireless bridge thing so i thought i'd leave it to you to fix. But this is my guess as to what is broken.

Cheers,
Lantis.

lantis1008:
fixed in 37e2f114e47e81aba07602b902984146defed483

PR #419, opened 2015-12-28
Update 011-arpnat.patch

lantis1008:
fix dependency on wireless_tools (iwconfig)
as suggested in #418 .

Not confident on this so please check.


ericpaulbishop:
About 30 seconds too late....

lantis1008:
lol no worries.

PR #431, opened 2016-01-06
i18n/Polish: add missing translations

obsy:


lantis1008:
Thanks mate, google translate sucks. I tried to tag you to ask if you could do it but you never responded.

PR #432, opened 2016-01-07
Fix QoS MinRTT setting for all languages that are not English-EN

lantis1008:
Any language that was not English-EN was unable to save a MinRTT setting.
Reported by user gatalon
http://www.gargoyle-router.com/phpbb/viewtopic.php?f=12&t=7934&p=34780#p34770


obsy:
BTW:
```
	if (classData[classIndex][4] == qosStr.YES) 
			{
				uci.set("qos_gargoyle",classId,"minRTT","Yes");
			}
```
IMO very ugly :) and "Yes" should be replaced by 1 (value as boolean, not string "Yes")

PR #435, opened 2016-01-11
Bump CC to 48220. Add DFS channels to GUI. Bring REGDB up to 20151214

lantis1008:
Bump Chaos Calmer to 48153.

Add DFS channels to the allowed list on the GUI to go with the changes made by @obsy in b38dec13995189fbef5d71053511377221475b56 .

Bring wireless-regdb up to date with the latest version as supplied by the official repository here:
http://git.kernel.org/cgit/linux/kernel/git/sforshee/wireless-regdb.git
If we are going to start allowing DFS, then this should be kept at the latest version.


obsy:
Also see https://dev.openwrt.org/changeset/48220 ( resync regulatory db with master-2015-12-14)

lantis1008:
Ah. That's annoying. Didn't realise they would do that so quickly after I suggested it. 
I'll update the pull.

lantis1008:
Reg Domain patch is now included in CC branch so it is now removed from /patches-generic

PR #440, opened 2016-01-20
Enforce DHCP assignments

nworbnhoj:
"Enforce DHCP assignments" replaces "Block MAC addresses assigned a
static IP that connect from a different IP" by using /tmp/dhcp.leases
rather than /etc/ethers to construct the iptables rule

See [forum discussion](http://www.gargoyle-router.com/phpbb/viewtopic.php?f=12&t=7937#p33723)


nworbnhoj:
I thought that the "why" was covered in the forum discussion?

Yes, the whitespace mess is annoying. I had a go at setting the git whitespace flagt but it made little difference. I will have another crack at it.

nworbnhoj:
Thanks - yep exactly.

Also, once we get the Known Devices incorporated then it will be relatively easy to "Block Unknown Devices" and get pretty good control of things (for example if a wifi password is passed on to "others")

lantis1008:
Ok I understand now. 
I must have skim read it the first time. I'm catching up on these things on my lunch breaks now :P

PR #442, opened 2016-01-22
i18n/Polish: fix translation

obsy:


nworbnhoj:
Thank you :-)

It would be good if we had list of people that we could msg when-ever we make an english i18n change and need translations. Maybe I should request for translators on the forum? Is there an easy way to diff all of the translations that are outstanding at the moment?

obsy:
Run ex ./i18n-scripts/comparitor.py English-EN German-DE and see difference between them.

lantis1008:
Obsy for Polish
Cappiewu for Chinese
Lyrz (only on forum) for French

That's the list I know of.

PR #443, opened 2016-01-23
fix call to renamed enforce_dhcp_assignments

nworbnhoj:
whoops - forgot to rename the call to function enforce_dhcp_assignments


ericpaulbishop:
I'm pretty sure I already fixed this earlier this morning in 517ec9c9c226cf54e9c05bdaa55cc37072140aa6

PR #448, opened 2016-01-30
Security improvements

nworbnhoj:
A number of security improvements are proposed:

1. Secure alternatives are recommended to the Gargoyle admin where applicable
2. The generic Gargoyle https private-key and certificate are deleted and regenerated on-demand
3. The Gargoyle admin is able to upload  and administer authorized ssh public-keys

there are more improvements planned...


ericpaulbishop:
I do really like enabling private keys for ssh. However, Two things, before I merge this:

(1) Https-only for the local interface is not recommended. The problem is that people will enable this and then they will get a certificate error. There is no way to avoid a cert error, because the IP is local. For remote access, yeah, if they're enabling that https is recommended, but because of this issue local https only is not recommended.

(2) Not all builds have the openssl utility. That may change, especially since I'm now looking into ditching the 4M routers, but for now it's there. Instead of removing the default certs, how about we check if openssl is present, and then, if it is, generate new ones upon enabling https? We can calculate md5sums or sha1sums of default certs to see if current certs are defaults.

nworbnhoj:
Agreed, there is no way to avoid the browser certificate error. However, I subscribe to https-everywhere and wonder if a better solution is to recommend https-only and then trigger a pop-up warning to educate the Gargoyle admin when they actually select it?
A number of forum users are clearly managing "inquisitive" hosts on their internal networks (setting static IP's, spoofing macs, resetting routers etc to get around quotas etc) - that's where we all start learning right! 

Sorry, I was not aware that openssl might be unavailable and will enhance this as you suggest. Thankyou.

BTW. In my travels I was bemused to observe that the `<select id=remote_web_protocol>` is set but never got! It seems that the function is not implemented in the GUI? Never heard anybody complain that it does not work - maybe no-one is crazy enough to try it. Should I fix it or ditch it?

nworbnhoj:
OK I think I am done here for a while. 

I also took a peek at the SSH server keys and they seem OK with 2048bit RSA and 1024bit DSA. The DSA is inadequate but it can stay for compatibility reasons. Also, there was some openWRT chatter a while back about the dubious merits of generating keys at first boot when there is a lack of entropy. https://dev.openwrt.org/ticket/9631 Probably, it would be better to replace them when required (as for uhttpd) but the check for default/original keys is more tricky.

PR #453, opened 2016-02-11
Enable USB modem for ar71xx.usb

nworbnhoj:
Thank you forum users @petbrau & n8b

CONFIG_PACKAGE_kmod-usb-net-cdc-ether=y


nworbnhoj:
Probably this should be added to other profiles as well. How do I choose which ones?

lantis1008:
Usually you would add to all except the smallest space routers. 
Be chase it's a module associated with usb I'd just do
Ar71xx.usb, usblarge, possibly the ath10k (I think they all have USB ports)
Mvebu

ericpaulbishop:
lantis is correct, please update as he suggests before I mege.

PR #455, opened 2016-02-13
Optimized german translation of adblock plugin

petbrau:
I suppose, the original translation was made with a translation software. So I modified the translation a little bit, to increase the comprehension.
I did not translate "Adblock", because this word is well know and the translation like "Werbeunterdrückung" or something else, sound strange.


lantis1008:
I appreciate this. Thank you.

Issue #461, opened 2016-02-17
Spectrum Analyser


obsy:
1) For old b/g routers, section "HT operation" does not exists, only "DS Parameter set: channel". See:
```
BSS e0:cb:4e:xx:xx:xx(on wlan0)
	TSF: 161492480580 usec (1d, 20:51:32)
	freq: 2462
	beacon interval: 100 TUs
	capability: ESS Privacy ShortSlotTime (0x0411)
	signal: -74.00 dBm
	last seen: 7010 ms ago
	Information elements from Probe Response frame:
	SSID: xxxxxxxxxxxx
	Supported rates: 1.0* 2.0* 5.5* 11.0* 18.0 24.0 36.0 54.0 
	DS Parameter set: channel 11
	TIM: DTIM Count 255 DTIM Period 1 Bitmap Control 0x0 Bitmap[0] 0x0
	ERP: Barker_Preamble_Mode
	ERP D4.0: Barker_Preamble_Mode
	RSN:	 * Version: 1
		 * Group cipher: CCMP
		 * Pairwise ciphers: CCMP
		 * Authentication suites: PSK
		 * Capabilities: 1-PTKSA-RC 1-GTKSA-RC (0x0000)
	Extended supported rates: 6.0 9.0 12.0 48.0 
	WPS:	 * Version: 1.0
		 * Wi-Fi Protected Setup State: 2 (Configured)
```
2) Special chars should be replaces:
ex:
```
	last seen: 370 ms ago
	Information elements from Probe Response frame:
	SSID: \x20Social WiFi
	Supported rates: 1.0* 2.0* 5.5* 11.0* 6.0 9.0 12.0 18.0 
	DS Parameter set: channel 8
```
\x20Social WiFi should be " SocialWiFI" (name with space)

lantis1008:
Issue 1 will be fix within 24 hours. 

Issue 2, as a limitation of the code, any spaces in SSID I remove. 
Are you saying that an SSID beginning with a space shows up as \x20?
So the plugin currently displays the special character as part of the name?

obsy:
2. Exacly. Just replace \x20 by space.

lantis1008:
I'll look into it. 
That starts down a slippery slope of Irving to handle all special characters. 

Putting a leading space may be possible. I don't remember if the issue was only with them partway through the string. 

Currently "wifi network" becomes "wifinetwork". Do you have  an opinion on this?

obsy:
[...]

PR #467, opened 2016-02-20
Fix non-ht routers in spectrum analyser

lantis1008:
fixes part 1 of #461


lantis1008:
sorry for the unclean pull. it should be fixed now.

PR #469, opened 2016-02-21
Add country selection option under System->Identification page

lantis1008:
Allows the user to select the country they are in for regulatory purposes. The default option ensures regulatory compliance in all countries and follows the standard implementation as per Openwrt.


ericpaulbishop:
So... I appreciate the work you did on this feature, and I agree that it would be a nice feature to have. In a vacuum, I would merge this.

However... I don't want it to be easy to switch away from the default limits on the radio frequencies. 3rd party firmware is, unfortunately, already in a bit of a gray area with regulatory bodies like the FCC. I REALLY want to avoid any such agency taking an interest in this project, and I would prefer to try to bend over backwards to comply with whatever regulations exist, whenever possible. 

For that reason, I don't want it to be easy to switch the allowed radio frequencies like this. Even though I really value your work, and would rather not have to do this, I am going to deny this pull request.

lantis1008:
I can appreciate that. 
I have a similar view and hence why I put the setting on a different page to the wireless section to try to obscure it.
But I understand your decision. No worries :)

nworbnhoj:
I also agree that Gargoyle should comply with regulations. No question.

However, I wonder if we can be a little smarter. What if we were to detect the country based on a ARIN whois lookup or one of the free IP Geolocation databases (probably 95%+ accuracy for country). Then we could offer the Gargoyle GUI user with only 2 options Default=World or Detected=country. Would this be an acceptable solution @Eric ?

lantis1008:
Doesn't work on routers that are not exposed to WAN

nworbnhoj:
If the device is not exposed to WAN then offer only default=world. 

The philosophy is to offer the user an additional option if and only if independent geolocation is possible. 

The user can accept the geolocated country if it is correct or stick with default=world. The user can only make an illegal wifi configuration if geolocation selects the wrong country (<5%) AND the user selects the wrong country AND the user selects a disallowed wifi channel. Surely an unlikely probability.

obsy:
I live in Poland. In Europe most countries allowed channel 12/13, and lot of people uses this channels. We must be able select country option, without this router has limited functionality.

obsy:
Ubiquiti also has country selection on first screen after hardware reset.

lantis1008:
in 1.9.x branch inquisitive user can do so on their own.

ericpaulbishop:
I actually like nworbnhoj's suggestion above, but let me add a few restrictions/caveats:

How about we do this on the initial firstboot page which sets password, language and timezone, and nowhere else, and also do it with a lookup to ARIN as suggested?

So, if there is no connection, it would just go to the default. If there is a connection, it would detect the country from the IP. Then there would be a message on that firstboot page: "Gargoyle has detected that you are in [YOUR_COUNTRY]. Do you wish to configure your wireless radio for use in [YOUR_COUNTRY] Yes/No (with default No). If they say yes, the country gets set, if they say no, it goes back to default.

This would mean that they have to manually confirm their country along with having it detected by the IP which adds another level of certainty, and it would also make it so that this isn't a parameter that can be changed after first configuration. I want to make it HARD to change this parameter, though there is a great deal of value here so long as it's set properly.

lantis1008:
That sounds doable. 
I'm heading away on a business trip today and I'm almost tempted to throw my router in my suitcase for the evenings. 

I'll certainly look at this style of implementation when I get back.

obsy:
+1 for me.

nworbnhoj:
If memory serves - the WAN connection is not initiated until immediately after the initial firstboot page so this does not quite work as is. Perhaps the initial firstboot page could be extended to include the basic WAN config and then a popup to confirm the auto-geo-located country before proceeding the the main Gargoyle GUI.

lantis1008:
Yeah I had that thought. 
I was thinking maybe the first time the user hits save on the basic config page we trigger it?
We already have a first boot flag in the system I believe. I've seen it in other bits of code.

ericpaulbishop:
Initially WAN network connection is via ethernet. If the ethernet cable is plugged in when the router boots for the first time, it should be there when the firstboot screen is displayed. We're already loading the available language packs on the firstboot screen by connecting to the internet and determining which language packages are available.

nworbnhoj:
Perhaps we need to be thinking about geo-locating the country on **first-connectivity** rather than **first-boot**. For example, my prior Gargoyle setup was via a USB 3G modem.

ericpaulbishop:
How about we put it on the first boot page if there is a connection. Then, if it has not been set, it shows up on Connection/Basic (in the form I describe: is this your country? yes/no) and then it disappears as soon as the user makes a selection, never to return.

lantis1008:
I'm going to put this on the back burner until I fix bridge configs. i plan to do that this weekend.

m45t3r:
Just a question, is it really necessary for this option to show only once?

I mean, says that I go to a trip somewhere and take my router with me. So now I need to either:
-  Reset my router to default settings.
-  Or find the config that disable this option, set it to false and do it again.

Considering that 99% of the cases Gargoyle is run as the main router in the house (and of course, 93% of all statistics are invented), I think that always showing the option World and <my current country> in Connection is ok, without need to all this heuristic of "first boot or first connection or first whatever" (and I am sure that we will get this wrong and some special case will not work).

There is only one case that I can think that maybe this will not work: if user is using VPN. In this case I would disable this option if user has a VPN configured.

lantis1008:
I've never seen it default to anything other than 00 World. But I've only tested 4 routers.

lantis1008:
That being said all my routers were bought in Australia so there could be a difference.

lantis1008:
As you can see from the discussion, changes have been requested to move the change in line with the vision of Gargoyle. I have not had time to do so. 

If anyone else wants to pick up this change they are welcome to do so. Otherwise I will get to it at some point in the future.

lantis1008:
Superseded by #752

PR #470, opened 2016-02-21
Update config

proper88:
Enable USB modem for ar71xx.usb
CONFIG_PACKAGE_kmod-usb-net-cdc-ether=y


proper88:
1.) Netgear WNDR3700v4
2.) because in
targets/ar71xx/profiles/usb
targets/ar71xx/profiles/usb_large
this modification is still included(only in targets/ar71xx/profiles/usb_large_nand missing)

lantis1008:
this is in reference to be6ac8c524d0a2c578862896e1b19172507fad6c and 04a7634b83e5bdf84c265cde0a6deadf7ee683f4

PR #481, opened 2016-02-28
Correct Connection List for large private networks

### LINKED ISSUE #452: The list of the current connections is incorrect.
I had a setup on my lan which was 192.168.0.1/24

Since I switched to a lan with address 192.168.0.1/21 The list of the current connections are incorrect because the lan address are displayed as WAN

The attached screenshot shows:
-in the first line the connection LAN->LAN which is reported as WAN->LAN
-in the second line the connection LAN->WAN (reported as expected, so fine)

I'm using 1.6.1 on TP-Link TL-WR1043N/ND v1

garg

nworbnhoj:
resolves #452


ericpaulbishop:
I had to revert this. It doesn't work. You can't assume one end of the connection is the router. Any IP in the router's subnet is fair game.

nworbnhoj:
Drat - there is the danger in writing code in the absence of a router. I will take another look at it.
Strange though, because I followed the notion of the existing logic and tweaked it lightly. Must have missed something (obviously).

PR #483, opened 2016-02-28
Fix git links and enable setting blank commit hash to checkout latest

lantis1008:
Git links didn't work for me unless they were in this format.
Added a quick IF statement so that if we set no commit hash number then we don't git checkout, which should leave it at the latest commit. This matches the existing SVN behaviour (although it was rarely used).


ericpaulbishop:
Oops... sorry about that. Thanks for pointing it out quickly.

PR #488, opened 2016-03-03
Enforce  DHCP assignments : typo.

roadhawk:
Enforce DHCP assignments : typo caused invalid iptables syntax.


nworbnhoj:
I have also tested this correction and it is valid.

Out of interest man iptables at http://linux.die.net/man/8/iptables lists a syntax which is invalid on TL-WDR3600

PR #492, opened 2016-03-03
standard version numbers for language files

petbrau:
I realized that the language files can not be selected, when using the date as a version number.

### DIFF
--- a/package/plugin-gargoyle-i18n-German-DE/Makefile
+++ b/package/plugin-gargoyle-i18n-German-DE/Makefile
@@ -1,8 +1,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=plugin-gargoyle-i18n-German-DE
-PKG_VERSION:=20160221
-PKG_RELEASE:=1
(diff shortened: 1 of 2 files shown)


obsy:
Selected where?

petbrau:
The Language files can be installed as plugin and are marked as installed in the plugin list. 
If you then switch to the system/language page, the Polish and the German language can not be activated. No problems with other language files.

obsy:
Hmm:

a

```
root@Gargoyle:~# opkg list_installed | grep i18n
gargoyle-i18n - 0.2.1-1
haserl-i18n - 0.9.26-1
plugin-gargoyle-i18n-English-EN - 1.0.0-1
plugin-gargoyle-i18n-Polish-PL - 20160228-1
```

ericpaulbishop:
obsy clearly shows that whatever issue you are having isn't the version number. If you find another reason you are experiencing a problem that can be replicated, I would be happy to merge a fix for that, but until then I'm closing the pull request.

PR #493, opened 2016-03-04
Fix patch 004

lantis1008:
The two chunks were failing. This is a quick update to reflect the latest files to be patched but i have not attempted a build with it yet.


ericpaulbishop:
Your modified patch did not apply cleanly. When you say you didn't do a build I didn't realize you hadn't even tried to apply the patch to the source.

In the future, please don't send pull requests that are untested (at all). Open an issue and attach code noting that it's untested, but I would prefer to avoid merging anything that is untested if at all possible.

PR #500, opened 2016-03-16
Use faster table methods

stopspazzing:
Faster and easier to use .createTHead() and row.insertCell() than to create a element and append to it. When deleting rows, use deleteRow(). See http://www.w3schools.com/jsref/dom_obj_table.asp for more improvements.

Don't merge quite yet, there is some more code I can edit to speed things up. Will be doing revisions.

### DIFF
--- a/package/gargoyle/files/www/js/table.js
+++ b/package/gargoyle/files/www/js/table.js
@@ -12,20 +12,19 @@ function createTable(columnNames, rowData, tableId, rowsAreRemovable, rowsAreMov
 	controlDocument = controlDocument == null ? document : controlDocument;
 
 	var newTable = controlDocument.createElement('table');
-	var thead = controlDocument.createElement('thead');
-	newTable.appendChild(thead);
+	var tableHead = newTable.createTHead();
 	var tableBody = controlDocument.createElement('tbody');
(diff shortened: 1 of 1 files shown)


d3fz:
Any news about this ?

Would these changes cause any problems if i test it on my own router ?

ispyisail:
@stopspazzing whats the status of this patch?

lantis1008:
I think this one should maybe just close. I don’t know that there are any performance gains to be had here. 
Plus, promised revisions have been abandoned an is now out of sync.

d3fz:
Agreed. 

I've also tested this locally and apparently it's not finished as it is, tables stopped being generated after applying the suggested changes.

lantis1008:
Hi @stopspazzing, do you think we could close this one please?

stopspazzing:
Closing it as requested. It is a faster method and cleaner looking in code, but not end of world critical speeds. It will be a side project for me after the theme is finished.

stopspazzing:
> Hi @stopspazzing, do you think we could close this one please?

Wow... so normal appending child's is faster it seems, just looks ugly and isn't proper way to do it. Weird.

https://jsbench.me/yokjapc2uk/1

PR #501, opened 2016-03-16
Fix basic.js and implement relayd for bridge configs

lantis1008:
- [ ] Basic.js should now behave correctly under all configuration types. 

- [ ] Bridge configs are now implemented using relayd. Router IP should be **DIFFERENT** to AP subnet. I haven't changed the translation files to reflect this yet.

Please thoroughly test this patch, i've done my best to make sure it is ready.


package/gargoyle/files/www/js/basic.js
```diff
@@ -661,7 +674,21 @@ function saveChanges()
...
+			{
+				preCommands = preCommands + "\nuci set network.bridgecfg=interface\nuci set network.wwan=interface\n";
+				uci.set("network", "bridgecfg", "proto", "relay");
+				uci.set("network", "bridgecfg", "network", "lan wwan");
+				
+				uci.set("network", "wwan", "proto", "dhcp");
+				
+				uci.set("dhcp", "lan", "ignore", "1");
+				
+				uci.set("firewall", "lan", "network", "lan wwan");
```
lantis1008: this line is no good. should be something like
```
uci.set("firewall", "@zone[0]", "network", "lan wwan");
```

Need to test whether @zone[0] works all the time, which i suspect it doesn't. May need a better way of identifying the correct config section.

package/gargoyle/files/www/js/basic.js
```diff
@@ -109,6 +109,17 @@ function saveChanges()
...
+		
+		//always remove bridge configs
+		preCommands = preCommands + "\nuci del network.bridgecfg\nuci commit\n";
+		uci.removeSection("network", "bridgecfg");
+		uciCompare.removeSection("network", "bridgecfg");
+		preCommands = preCommands + "\nuci del network.wwan\nuci commit\n";
+		uci.removeSection("network", "wwan");
+		uciCompare.removeSection("network", "wwan");
+		uci.set("dhcp", "lan", "ignore", "0");
+		uci.set("firewall", "lan", "network", "lan");
```
lantis1008: consequently this line also sucks.

lantis1008:
Fixes #504 . Still needs fixing on the firewall script before it is merged. Apologies.

lantis1008:
Firewall rules should now work properly.

ericpaulbishop:
I've tested this and, I couldn't find any issues. Good job!

I have this nagging suspicion that I've missed something, given how complex this is, and I'm not sure I want to do a release quite yet (though we're overdue for one), but this looks very, very good. Thank you.

PR #506, opened 2016-03-26
Additional 302 loop fix.

jaf0:
* when a theme does not contain `images/favicon.png` file, attempts to
  access the .png cause a 302 Found to overview.sh in the relative path.

* added leading slash to firstboot.sh for consistency.


jaf0:
I guess the alternative is to symlink favicon.png to ../../Gargoyle/images/favicon.png

Not sure which is truly better.

jaf0:
i have another tweak; i'll close this, and make branches...

PR #514, opened 2016-04-06
typo: comma missing

obsy:


lantis1008:
Will this fix the list option problem or do I still need to look at that?

lantis1008:
Ah I see that Eric has already fixed my error. 

Thanks to both of you.

obsy:
Yes, with this small fix  is ok.

obsy:
@ericpaulbishop  can we remove completely arp-nat patch?

Issue #537, opened 2016-05-15
CDC-NCM ethernet lan and wan bridge


luki92:
When you select DHCP WiFi client protocol, there is an option to bridge WAN ethernet port with LAN.
When you select CDC NCM protocol, this option should be also available.

Actually this option is not visible, but would be very useful; especially on devices with only 1 LAN port and 1 WAN port.

I have also a specific (OpenWRT related) issue, that my modem is not recognised by OpenWRT. This is because it reports just "E3372" as a model. I had to add: `[ -z "$protocol" ] && protocol="huawei"` to lib/netifd/proto/ncm.sh to get it fixed.

lantis1008:
@obsy I'll fix the bridge port if you could please handle the modem? That's your expertise :) thanks

obsy:
@luki92: show me

cat /sys/kernel/debug/usb/devices

luki92:
T:  Bus=02 Lev=01 Prnt=01 Port=00 Cnt=01 Dev#=  7 Spd=480  MxCh= 0
D:  Ver= 2.10 Cls=00(>ifc ) Sub=00 Prot=00 MxPS=64 #Cfgs=  1
P:  Vendor=12d1 ProdID=1506 Rev= 1.02
S:  Manufacturer=MOBILE
S:  Product=MOBILE
C:* #Ifs= 3 Cfg#= 1 Atr=80 MxPwr=  2mA
I:* If#= 0 Alt= 0 #EPs= 2 Cls=ff(vend.) Sub=03 Prot=12 Driver=option
E:  Ad=81(I) Atr=02(Bulk) MxPS= 512 Ivl=0ms
E:  Ad=01(O) Atr=02(Bulk) MxPS= 512 Ivl=0ms
I:  If#= 1 Alt= 0 #EPs= 1 Cls=ff(vend.) Sub=03 Prot=16 Driver=huawei_cdc_ncm
E:  Ad=83(I) Atr=03(Int.) MxPS=  16 Ivl=2ms
I:* If#= 1 Alt= 1 #EPs= 3 Cls=ff(vend.) Sub=03 Prot=16 Driver=huawei_cdc_ncm
E:  Ad=83(I) Atr=03(Int.) MxPS=  16 Ivl=2ms
E:  Ad=82(I) Atr=02(Bulk) MxPS= 512 Ivl=0ms
E:  Ad=02(O) Atr=02(Bulk) MxPS= 512 Ivl=0ms
[...]

PR #538, opened 2016-05-16
ncm: alllow bridge wan to lan when NCM is used

obsy:
See https://github.com/ericpaulbishop/gargoyle/issues/537


lantis1008:
Very kind of you. Thanks!

PR #540, opened 2016-05-26
Samba fixes. New language Slovak-SK. Add temperature script. Bump Spectrum Analyser.

lantis1008:
- The "valid users =" clause breaks the guest access.
By removing it, all expected behaviour is retained.
If the user wants write access, they need to map the drive with authentication credentials. Otherwise they will just be allowed to access the drive without a prompt with the permissions given to the guest (r/rw). I have only commented the line out instead of removing it entirely to make the commit easier to revert in-case there is a usage case that i have not accounted for that does not behave correctly without the "valid users=" clause.

- Preserve samba passwords across backups.

- Introduce new language pack Slovak-SK.

- Remove the temperature detection from overview page into a new external script. Allows for quicker and easier maintaining of existing and future models that support hwmon.
- Add WRT1900ACS to the temperature detection list.

- Update Spectrum Analyser plugin to 1.0.4. Fixes routers which don't strictly follow 802.11 IEEE standards for broadcasting their channel data


nworbnhoj:
Well done! That has been a PITA for ever.

lantis1008:
Big hint from forum user "thekk". Still had to do a lot of research and testing to confirm his conclusions but they turned out pretty spot on.

PR #550, opened 2016-06-18
Generate Gargoyle images for TP-Link TL-WR841 v10 and v11 routers

magyarsz:


lantis1008:
While there is no harm leaving it in there, this won't produce an image for v11. It has not been back ported to the openwrt version we are currently sitting at as far  as I am aware. 

Have you tried this @magyarsz ?

magyarsz:
Yeah, Of course I checked before I sent.

The Gargoyle based on the OpenWRT CC branch commit eadf19c0b43d2f75f196ea8d875a08c7c348530c (2016-04-27) now.
The support for routers above backported with commit 64e116779c0f7da6d98068b8e7c50f528c8a91f2 (2016-04-25).

So these devices are supported.

lantis1008:
Yep my apologies. I've not been looking at the commits lately have been offline.
This commit is fine.

PR #552, opened 2016-06-30
missing ;

stopspazzing:
missing ; on some end of lines.


ericpaulbishop:
This really isn't necessary.

In javascript the semicolons aren't necessary unless the two statements are on the same line. They're optional. Because I'm used to coding in C where the semicolons are required I have often put them in out of force of habit, but because they aren't required, there aren't any problems where they are absent. All of this goes through a js minimizer anyway, so it really doesn't matter.

You could make the argument that for purposes of having a consistent style it should get fixed... the problem being that this is much more widespread in the codebase than just what you adjust in this pull request. I'm going to close the pull request as-is because it doesn't really make things consistent, and I'm reluctant to head down the rabbit-hole of trying to do this everywhere.

PR #557, opened 2016-07-10
Fix strings for new bridge requirements

lantis1008:


petbrau:
the german translation should be: 
basicS.BrNote="(darf nicht im Subnetz des AP sein)";

PR #560, opened 2016-07-13
fix 30-hostname_dns hotplug script to work correctly with CC

magyarsz:
This issue has been reported in the following post on the forum:
http://www.gargoyle-router.com/phpbb/viewtopic.php?f=3&t=8890&start=10#p39287


ericpaulbishop:
This is going to wipe out all hosts in the hosts file when it runs. This "fix" will cause a problem bigger than the one it solves.

magyarsz:
Yes, I know about it. In this case, I'm still working on this in the future.

magyarsz:
I think I found a more acceptable solution.

PR #564, opened 2016-07-20
Add new target default_small

### LINKED ISSUE #524: [Proposal] Create a new target for 4MB Flash/no USB routers
I was playing with some compile options in OpenWRT CC-branch, and got a build that works stable in my 4MB Flash router (TL-WR941ND). Removing all support for USB, removing debug information and stripping kernel from unecessary code (dropping support for external modules), I could fit a fairly complete build of OpenWRT-CC with LuCI+luci-app-sqm+luci-app-upnp and still have a 512KB `/overlay` partition.

So I tried to port the same configuration file to Gargoyle, and this is the result: [config.zip](https://github.com/ericpaulbishop/gargoyle/files/233184/config.zip)

I could compile a Gargoyle-CC with support for miniupnpd (that was removed from CC because of the size), include all simple plugins (plugins smaller than 5KB, like webshell and initd) and still have a 512KB '/overlay' partition, that is bigger than the 320KB partition from Gargoyle BB. However some things didn't work, like WAN IP in Status page was reporting the LAN IP from my router, instead of the real IP (I am using PPPoE). Most notably, neither *B/W Distribution* and *QoS Distribution* graphics work, probably because I removed all debug support from kernel.

Anyway, my proposal here is to create a new target, say `4mb-nousb` or something, specific to routers with 4MB of Flash and no USB support, that excludes support for some modules to increase the available space, maybe bringing back miniupnpd in official builds and having more space to new features in the future. I think that removing all USB modules and stripping the resulting kernel should be sufficient. Since those routers don't have USB support (officially, anyway), there is no need to care about these builds being non-compatible with PPTP or OpenVPN plugins (that require additional modules), since it wouldn't fit anyway.

lantis1008:
Supports nothing. Gargoyle with no frills and no debugging. If this doesn't work, get out your wallet and buy a better router.

Closes #524


lantis1008:
No longer participating in this issue.

PR #565, opened 2016-07-21
Added default_small profile

### LINKED ISSUE #524: [Proposal] Create a new target for 4MB Flash/no USB routers

m45t3r:
default_small is similar to default, however it disables debugging features in kernel and strip kernel image and libraries to reduce the size. The result is that small routers with 4MB of Flash will have more spare space, allowing miniupnp package to be reactivated too (bringing UPnP support again, missing since Gargoyle 1.9).

The consequence of stripping kernel image and libraries is that the resulting firmware will not be compatible with packages from OpenWRT itself. Only the packages compiled with the firmware will be compatible. However this should not be a problem, all targets of this profile are routers with small Flash AND no USB support. So it generally impossible to install too many things on it.

This commit reactivates miniupnp for default profile too, since the target routers in this profile should be big enough to fit miniupnp.

Closes #524.


m45t3r:
For TL-WR941NDv3, this profile results in `/overlay` of 448KB in size, contrary of 320KB from the current `default` profile. And this profile includes `miniupnpd`, removed from official builds because of size.

m45t3r:
@lantis1008 suggested in issue #524 that it may be desirable to disable USB related plugins from this PR. It would be something like the patch below applied in `targets/ar71xx/profiles/default_small/config` file:

```
diff --git a/config_default_small b/config_default_small_nousb
index 0d82452..3597775 100644
--- a/config_default_small
+++ b/config_default_small_nousb
@@ -1398,8 +1398,8 @@ CONFIG_PACKAGE_plugin-gargoyle-i18n-Slovak-SK=m
 CONFIG_PACKAGE_plugin-gargoyle-i18n-Spanish-ES=m
 CONFIG_PACKAGE_plugin-gargoyle-initd=m
 CONFIG_PACKAGE_plugin-gargoyle-logread=m
-CONFIG_PACKAGE_plugin-gargoyle-minidlna=m
-CONFIG_PACKAGE_plugin-gargoyle-openvpn=m
+# CONFIG_PACKAGE_plugin-gargoyle-minidlna is not set
+# CONFIG_PACKAGE_plugin-gargoyle-openvpn is not set
 CONFIG_PACKAGE_plugin-gargoyle-ping-watchdog=m
 CONFIG_PACKAGE_plugin-gargoyle-pptp=m
 CONFIG_PACKAGE_plugin-gargoyle-spectrum-analyser=m
@@ -1411,10 +1411,10 @@ CONFIG_PACKAGE_plugin-gargoyle-theme-green=m
 CONFIG_PACKAGE_plugin-gargoyle-theme-horchata=m
 CONFIG_PACKAGE_plugin-gargoyle-theme-light=m
 CONFIG_PACKAGE_plugin-gargoyle-theme-xeye=m
-CONFIG_PACKAGE_plugin-gargoyle-tor=m
-CONFIG_PACKAGE_plugin-gargoyle-usb-printer=m
-CONFIG_PACKAGE_plugin-gargoyle-usb-storage=m
-CONFIG_PACKAGE_plugin-gargoyle-webcam=m
+# CONFIG_PACKAGE_plugin-gargoyle-tor is not set
+# CONFIG_PACKAGE_plugin-gargoyle-usb-printer is not set
+# CONFIG_PACKAGE_plugin-gargoyle-usb-storage is not set
+# CONFIG_PACKAGE_plugin-gargoyle-webcam is not set
 CONFIG_PACKAGE_plugin-gargoyle-webshell=m
 CONFIG_PACKAGE_plugin-gargoyle-wifi-schedule=m

```

ericpaulbishop:
I've updated the ar71xx/default profile, which is *already* what you would call the default small profile -- this profile is the fallback and it's mostly only 4M routers in here. 8M routers should probably be moved to ar71xx/usb (even if they have no usb port), as the default profile really is now optimized to make 4M routers work. If there are some routers that should be moved over to ar71xx/usb, based on having 8M (even if there is no usb port), please send a pull request for that.

I've added part, but not all of what you suggest here. I'm setting the options you suggest to strip out debugging info, but I'm not going to toggle CONFIG_STRIP_KERNEL_EXPORTS or CONFIG_USE_MKLIBS, which would mean that most if not all packages that you try to install on top of the original build won't work. My preference is not to re-enable miniupnpd and to keep these two options disabled to allow the installation of other packages if desired.

By the way, when comparing profiles you may find the config_diff script located in the dev-utils dir useful. If you run it on two config files it will show you the differences in packages and options enabled, regardless of the original order in the config file (where if you use regular unix diff the order of the options is going to matter and you'll see more differences than may really exist if only the order changes).

m45t3r:
Could we have at least CONFIG_STRIP_KERNEL_EXPORTS? This way, every package that is not a kernel module should work (this may confuse users, though).

And well, bringing back miniupnpd was really my objective in this PR. Using Gargoyle vanilla without it is really bad for home routers.

m45t3r:
JFYI, enabling only CONFIG_STRIP_KERNEL_EXPORTS should work with Tor/OpenVPN and other plugins from Gargoyle itself, it wouldn't work with kmod packages from OpenWRT-CC.

ericpaulbishop:
Actually.. yes, you do have a good point. Most packages don't really require kernel modules, so enabling that only should be a decent compromise for 4M routers.

ericpaulbishop:
Regarding miniupnpd... that is definitely not coming back for 4M routers. Already I've seen instability there because of how low on space they are. Freeing up more space is good, but, no, miniupnpd is gone.

m45t3r:
Ok, fair enough.

PR #568, opened 2016-07-28
fix band display issue if a client is connected in WDS mode

magyarsz:
_Symptom:_
If a client is connected in WDS mode to the another (main) router, the main router is always be displayed that the client is connected on the 5GHz band.


ispyisail:
Is there a reason why Eric hasn't pushed this patch?

ericpaulbishop:
Not a good one -- now merged. I remember going over several patches when
this was submitted that had some issues and for some reason remember this
one looking like it needed some work before merging, but it's only one line
and it's an important fix. I really should have merged it a while ago, and
I just did so.

PR #584, opened 2016-12-29
Last GUI changes

lantis1008:
All files (i think) have now been converted, excluding:
- firstboot.i18n.sh
- email plugin

Only 1 theme has been converted so far (not included in any changes thus far)
Waiting to bulk load them.


lantis1008:
@ericpaulbishop
With these changes largely complete, I have noticed an increasing amount of input elements that like to escape their parent containers.

Issue #587, opened 2017-01-03
Connecting USB printer not providing Product/SerialNumber tag fails


odenninger:
I connected a Lexmark Optra E laser printer (using a parallel port to USB adapter) to my router (Gargoyle Version:1.9.1/Model:TP-Link TL-WDR4300 v1).

As expected the printer is recognized, see `dmesg` output:
```
usb 1-1.1: new full-speed USB device number 6 using ehci-platform
usblp 1-1.1:1.0: usblp0: USB Bidirectional printer dev 6 if 0 alt 1 proto 2 vid 0x050D pid 0x0002

```
However, UI shows no connected USB printer.

I recognized that the printer configuration script `configure_printer.sh` doesn't handle my printer as it provides neither Product nor SerialNumber tag. The script tries to generate a printer name from Product tag or as fallback from SerialNumber tag. None of these tags is provided by my device, see `cat /sys/kernel/debug/usb/devices` output:
```
T:  Bus=01 Lev=02 Prnt=02 Port=00 Cnt=01 Dev#=  6 Spd=12   MxCh= 0
D:  Ver= 1.00 Cls=00(>ifc ) Sub=00 Prot=00 MxPS= 8 #Cfgs=  1
P:  Vendor=050d ProdID=0002 Rev= 1.04
C:* #Ifs= 1 Cfg#= 1 Atr=80 MxPwr= 98mA
I:  If#= 0 Alt= 0 #EPs= 1 Cls=07(print) Sub=01 Prot=01 Driver=usblp
E:  Ad=01(O) Atr=02(Bulk) MxPS=  64 Ivl=0ms
I:* If#= 0 Alt= 1 #EPs= 2 Cls=07(print) Sub=01 Prot=02 Driver=usblp
E:  Ad=01(O) Atr=02(Bulk) MxPS=  64 Ivl=0ms
E:  Ad=82(I) Atr=02(Bulk) MxPS=  64 Ivl=0ms
I:  If#= 0 Alt= 2 #EPs= 3 Cls=ff(vend.) Sub=00 Prot=ff Driver=usblp
E:  Ad=01(O) Atr=02(Bulk) MxPS=  64 Ivl=0ms
E:  Ad=82(I) Atr=02(Bulk) MxPS=  64 Ivl=0ms
E:  Ad=83(I) Atr=03(Int.) MxPS=   4 Ivl=1ms
```

[...]

PR #588, opened 2017-01-03
plugin usb-printer: Additional fallback to generate USB printer name

odenninger:
If an USB printer neither provides Product nor SerialNumber tag, the
printer configuration scripts fail to generate a printer name and exits.
This change adds an additional fallback combining Vendor, ProdID, and
Rev tag to generate a printer name. These tags should be available for
all USB devices.

This change fixes ericpaulbishop/gargoyle#587


obsy:
+1

odenninger:
Missed to test disconnecting a printer. Create new pull request #589 with fix.

PR #590, opened 2017-01-14
i18n/Polish: add missing translation

obsy:


package/plugin-gargoyle-i18n-Polish-PL/files/www/i18n/Polish-PL/time.js
```diff
@@ -1,3 +1,4 @@
+
```
lantis1008: Is the BOM preserved through this?

PR #596, opened 2017-01-23
Fixes to Bridge Config + Guest Network + Fix large graphs

lantis1008:
- Make bridge configs clearer by rearranging elements
- Fix bridge logic causing displayed settings to not match configured settings
- Display the Relay IP on the overview page so people can find it without digging in the console
- Disable Guest Network on b43 devices
- Fix large graphs

I feel like basic.js is becoming more and more cobbled together and may need a complete rewrite. I think that is mostly my fault. Low priority, but definitely standing in the way of IPv6.

### DIFF
--- a/package/gargoyle/files/www/bandwidth.sh
+++ b/package/gargoyle/files/www/bandwidth.sh
@@ -103,17 +103,17 @@
 
 				<div id="bandwidth_graphs" class="row">
 					<h1 class="page-header"><%~ BGrSect %></h1>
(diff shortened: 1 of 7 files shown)


lantis1008:
@ericpaulbishop ,
Please take a look at the final commit in regards to Graphs. There has been much feedback on the forums regarding the graphs being too big. I think this change makes them look much better at all sizes. Please test it out before accepting.

PR #598, opened 2017-01-27
compile fixes necessary to fix qos compile error

joequant:
These changes are necessary to avoid compile errors in qos.


ericpaulbishop:
qosmon gets compiled with the ONLYBG flag which means that ncurses should not be necessary.

PR #599, opened 2017-01-29
fix include of iproute

joequant:
The current qos makefile matches two directories for iproute causing
multiple define errors.


ericpaulbishop:
I just committed an update that should address the issue I suspect you're encountering, but I have to reject your pull request.

The way I had it was a bit of a hack I admit, but your pull request makes the situation worse, not better. Some builds (the ones for larger routers with more space) use iproute2-full, others use iproute2-tiny and whichever is used, qos needs to link to it. Your change would kill some builds, those with iproute2-full.

The update I just added should dynamically check what is present, and use iproute2-full if it exists, otherwise iproute2-tiny.

PR #601, opened 2017-01-30
Add kibosh.net dynamic dns service support

kiboshdev:
We run a DNS filtering service, and have added support for updating this service to ddns_providers.conf so that it can be supported in Gargoyle. Please consider merging this pull request and adding support for our service.

Thank you!


lantis1008:
The password is always the same?

kiboshdev:
Yes. The usernames we use are not easy to guess. This way users only need to remember and enter one piece of information instead of two.

PR #630, opened 2017-05-14
Collapse menus + RegDB update

lantis1008:
This PR has 3 main changes
1. Reg Database update to latest via patch. Some countries have had more spectrum opened up and higher transmission powers enabled.
2. Make Gargoyle menu system collapsible
3. Enable collapsible menus on default theme

The previous implementation of the menu system meant that if you clicked one of the major menu items, it would take you to the first item in that sub category e.g. Clicking on Connection took you to Connection->Basic
This new set of changes makes the major headings expand and collapse to reveal their sub-menus. This change is particularly useful on mobile devices where the number of clicks required to get to a particular page was increased.
This has been implemented by making the links into "dummy anchors" i.e. "#" and the links all return false with their onclick event, meaning the URL never changes to include the dummy anchor, reducing confusion.
If a major heading has zero sub-menus, the old behavour is retained i.e. clicking on it will take you directly to it's linked page. A good example of this is the logout button.

As part of this change i also fixed the tab indentation of the entire file which was previously destroyed, and implemented proper tab indentation of the resulting HTML once passed through haserl. This makes the source code in the browser much easier to navigate.
Unfortunately due to this, the diff is a bit all over the place, so if you would like me to step through the changes in more detail i can arrange this. A good example of the type of change is found at Line 560 of the header_footer file.

Individual themes can turn on and off the collapsing behaviour by including a single (empty) txt file within the theme with the name `collapseMenus.txt`. This adds minimal (none?) overhead to the theme size.

A demonstration of the new behaviour can be seen [here](https://www.youtube.com/watch?v=8_Y7mh8t9AE)

### DIFF
--- a/package/gargoyle/files/www/js/common.js
+++ b/package/gargoyle/files/www/js/common.js
@@ -9,6 +9,8 @@
 var UI=new Object(); //part of i18n
 var TiZ=new Object(); //i18n timezones
 var TSort_Classes = new Array ('odd', 'even'); // table sorting zebra row support
+var menuState = []; //nav Menu State variable
+var navTimer; //timeout for restoring menu
 
 window.onresize = function onresize()
 {
--- a/package/gargoyle/files/www/js/common.js
+++ b/package/gargoyle/files/www/js/common.js
@@ -2805,3 +2805,46 @@ function togglePopup(popElement)
 	}
 	descendant[0].classList.toggle("show");
 }
+
+//Part of collapsible menus scripts
+function uncollapseNavThis(menuElement)
+{
+	var parentSidebar = menuElement.parentElement.parentElement;
+	collapseNavOthers(parentSidebar);
+	menuElement.parentElement.classList.add("active");
+	var siblings = menuElement.parentElement.childNodes;
+	for (var x = 0; x < siblings.length; x++)
+	{
+		if (siblings[x].nodeName == "UL")
+		{
+			siblings[x].classList.add("active");
+		}
+	}
+}
+
+function collapseNavOthers(parentSidebar)
+{
+	var descendant = parentSidebar.childNodes;
+	if (descendant.length < 1)
+	{
+		return null;
+	}
+	else
+	{
+		for(var x = 0; x < descendant.length; x++)
+		{
+			if(descendant[x].nodeName == "LI")
+			{
+				descendant[x].classList.remove("active");
+				var seconddescendant = descendant[x].childNodes;
+				for(var y = 0; y < seconddescendant.length; y++)
+				{
+					if(seconddescendant[y].nodeName == "UL")
+					{
+						seconddescendant[y].classList.remove("active");
+					}
+				}
+			}
+		}
+	}
+}
--- a/package/gargoyle/files/www/js/common.js
+++ b/package/gargoyle/files/www/js/common.js
@@ -2845,3 +2845,56 @@ function collapseNavOthers(navElement)
 		}
 	}
 }
+
+function storeNavState()
+{
+	var selectorStart = "nav_MAJ";
+	var selectorEnd = "_MIN";
+	var x = 1;
+	var y = 0;
+	var selectorStr;
+	var elem;
+	var elemSub;
+
+	function createSelectorString(selectorStart,x,selectorEnd,y)
+	{
+		selectorString = selectorStart + ("0" + x).slice(-2) + selectorEnd + ("0" + y).slice(-2);
+		return selectorString;
+	}
+
+	selectorStr = createSelectorString(selectorStart,x,selectorEnd,y);
+	elem = document.getElementById(selectorStr);
+	while(elem != null)
+	{
+		menuState.push([elem.id,elem.className])
+		y++;
+		selectorStr = createSelectorString(selectorStart,x,selectorEnd,y);
+		elemSub = document.getElementById(selectorStr);
+		while(elemSub != null)
+		{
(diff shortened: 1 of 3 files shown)


ericpaulbishop:
This sounds really good. I'm testing it now.

ericpaulbishop:
I like the ability to navigate the menu without actually visiting the link. 

However, what I don't like is that the way you've implemented this, (1) you have to click on a section to see what's in it, which means clicking doesn't actually open the section, and thus clicking doesn't do what you would expect and (2) when you do click and open a different section you can't see where you are at that moment anymore. I kind of expect to see the menu indicate what section/page I'm currently at.

One way to fix this might be if I didn't have to click at all, but rather if hovering over the button would expand it, and the menu then returned to the current section after I move the cursor away.

I'm not going to merge this as-is (though I cherry-picked the regdb update), and I'm very open to something *like* this, just not as-is due to the usabiltiy issues listed above. I'm going to leave the pull request open so that if you want you can add more commits based on my concerns, and I will  re-evaluate.

lantis1008:
Thanks for the feedback. 
I'll revisit the useability. 

My only concern with hover actions is how they translate to mobile devices (you can't!). I'll do some tests to see how it behaves. 
I'll also look at rebounding the menu back to its current state.

d3fz:
Would this approach resolve such problem ?

http://uxmovement.com/navigation/how-to-simplify-a-large-navigation-menu/

lantis1008:
I'm looking at whether using mouseover events is reliable and doesn't break interface. 
Currently that's a no from my opinion. It makes the whole thing very jarring to use, not better. 

To overcome this you'd need to apply smoothing to the menu opening. 

There also needs to be a time delay function to reset the menu back to where you currently are.

tapper82:
Hi the mouseover events  can be a problem to screenreaders, But thay can work if dun rite. I am not shure about what is the rite way tho. I will giv feedback as soon as there is a build to test. One way that works that i know of is in Luci using the Material theme. EG wen I press enter on system up pops all the links under system. If i press enter again thay get hidden. The Bootstrap theme does not do that, it shows all available links.

lantis1008:
Hi @ericpaulbishop 

Please see/test the latest changes to this PR.
I have not done a build test, but have tested the code on a live router.

Changes:
- Nav elements now have an ID assigned to them making them easier to find/manipulate in functions e.g. "Status" is the first major nav item so it is "nav_MAJ01_MIN00". "Overview" is the first minor nav item under Status so it is "nav_MAJ01_MIN01" etc.
- On pageload, the nav state for that page is stored in a variable. I don't forsee this being a problem, but if it is it could be coded into the header/footer program itself
- If the user moves their mouse away from the menu for > 1.2s, the menu will restore to it's original state. This action is cancelled if the user puts their mouse back onto the menu before the timer lapses.
- As discussed, the "onclick" behaviour of the menu has now been changed to "onmouseover". I don't agree with this change and personally think it feels very clunky. Please spend a bit of time using this particular part and let me know your thoughts. It could maybe be solved by a slow transition applied to the menu, although i couldn't get this to work properly.

Please let me know of any feedback.
@tapper82 the onmouseover events should be ok for a screen reader. I navigated the menu using keyboard only and it operated as i expected.

lantis1008:
probably needs a onclick = "return false;" if staying with onmouseover.

ericpaulbishop:
Hmm... I see what you mean about the limitations of mouseover. 

I'll admit, the onclick interface DOES work really well for mobile devices, it's just really non-intuitive for desktop users. Maybe the best thing to do is implement this with onclick, but make that only active at all for mobile users -- so for now, Desktop users get the old behavior, while mobile users get the new behavior with onclick.

lantis1008:
I'm unsure how to achieve this without generating two Nav bars. 
One to be displayed at desktop resolutions, and the other to be displayed on mobile. 

Can you be swayed for allowing the desktop case new behaviour?
I agree that it is *strange* at first given what we are used to, but like most change, we can adapt with time.

My current thoughts are to return to implementation 1, with the menu returning back to its home state after the delay added in (you're 100% right about this one).

ericpaulbishop:
I'm pretty sure it's not only possible, but I just did it. See ef66f13. 

 I tested the javascript though, and it seems to be working.

I haven't merged it into master yet because I made some changes to gargoyle_header_footer I haven't tested yet (removing things that don't need to be there, mostly), and I haven't tested that it compiles ok/ that it's typo free yet. I'm doing a build to test that now.

lantis1008:
Cool solution. 
This is why you're in charge.

lantis1008:
@ericpaulbishop
How will this new JavaScript implementation go with a theme that doesn't provide the collapseMenus.txt in its theme? Will it fail gracefully?

I might look at a small rewrite to move to UCI variables in the global config now that the implementation is largely in JavaScript

ericpaulbishop:
Ah, excellent point, thanks for pointing that out.

From what I saw, I think it probably does fail gracefully but I haven't tested it and even if it's ok now if we change something further it might not. I think the best solution  is to add a javascript flag that gets written by gargoyle_header_footer, and then tested before calling the javascript that handles the collapsible menus.

ericpaulbishop:
Just committed a fix for this, and in doing so added a feature: theme-specific javascript. If a theme.js file is present in the theme directory, it gets automatically included by gargoyle_header_footer. I moved the call to the initialization code out of common.js and into the theme.js file for the default Gargoyle theme.

lantis1008:
Very cool. Went from a nice to have addition to a big value add.

PR #642, opened 2017-07-25
Fast fix for board wr8305rt, wan macaddr variable must ALWAYS be set.

mysticall:
After the initial setup, the gargoyle removes the mac address of the wan interface and after restart the wan port changes the mac address.


mysticall:
Please add and this patch:
https://github.com/mysticall/gargoyle/blob/master/spm/017-ramips-button-handlers.diff
Otherwise, the reset button does not work.

PR #644, opened 2017-08-07
Fix bug introduced by 1280cd6

lantis1008:


lantis1008:
Not run tested, but makes sense to me

crackself:
this fix make a new bug: could't update firmware via web upgrade.

lantis1008:
You can do so via the command line. 
I will look at where else this patch touched.

crackself:
could you show that command line,thanks.

lantis1008:
`sysupgrade /path/to/sysupgrade.bin`
Add -n option to not preserve settings. 
Move bin file to /tmp/ first

lantis1008:
Nothing obviously wrong with code. 
Can you please provide errors System logs as you did with the previous issue?

crackself:
ooh,find out the reason,this cause by the old firmware cache.after flash to a stable edition then try the new build firmware,it run fine.

lantis1008:
Great. No worries then.

Issue #643, opened 2017-07-29
The lastest update seems faile to save chages of wireless setting


crackself:
update code to commit 14a8cfaa75794d17f39744ae9c169dc00113823d and build(FULL_BUILD) for netgear 4300,it can not save change to wifi password.
![2017-07-29 08 44 12](https://user-images.githubusercontent.com/16448332/28740692-a796e67a-743a-11e7-9c05-7c09b67abfb6.png)

crackself:
this problem still exists on commit 10f4ee6383ecb34e60af98646c0bd66ea63df56f

lantis1008:
I'm not seeing the same issue. Can you please provide more details. 
Or even a system log?

crackself:
don't know how to output kernel log,here'ssystem log get from systemlog plugin,it seems many rule to do with dnsmasq(I have nothing change about it):
`Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 95.142.171.235#53 for domain bbs
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 95.211.32.162#53 for domain bbs
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 66.244.95.20#53 for domain bbs
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 95.142.171.235#53 for domain dyn
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 95.211.32.162#53 for domain dyn
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 66.244.95.20#53 for domain dyn
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 95.142.171.235#53 for domain parody
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 95.211.32.162#53 for domain parody
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 66.244.95.20#53 for domain parody
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 95.142.171.235#53 for domain glue
Sun Aug  6 13:28:27 2017 daemon.info dnsmasq[4387]: using nameserver 95.211.32.162#53 for domain glue
[...]

PR #649, opened 2017-08-22
Adjust CLEAN_IPKG_PARTIAL behaviour

lantis1008:
Now leaves the `.control` files for packages with names like `plugin-gargoyle-*` in-situ.
- Fixes blank entries on Languages.sh page (Closes second half of #643)
- Adds additional information to Plugins.sh page for pre-installed plugins

### DIFF
--- a/patches-generic/021-clean-ipkg-partial.patch
+++ b/patches-generic/021-clean-ipkg-partial.patch
@@ -5,7 +5,7 @@
  	rm -f $(TARGET_DIR)/usr/lib/opkg/info/*.postinst
  	$(if $(CONFIG_CLEAN_IPKG),rm -rf $(TARGET_DIR)/usr/lib/opkg)
 +	$(if $(CONFIG_CLEAN_IPKG_PARTIAL),rm -rf $(TARGET_DIR)/usr/lib/opkg/info/*.postinst)
-+	$(if $(CONFIG_CLEAN_IPKG_PARTIAL),rm -rf $(TARGET_DIR)/usr/lib/opkg/info/*.control)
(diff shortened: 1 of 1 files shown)


d3fz:
@lantis1008 ,

Could you please apply this fix to 1.11.X?

lantis1008:
@d3fz  it looks like it is already doing so:
https://github.com/ericpaulbishop/gargoyle/blob/base_on_openwrt_remerge/patches-generic/021-clean-ipkg-partial.patch#L10

Are you seeing bad behaviour from this new version of the patch?

d3fz:
>Are you seeing bad behaviour from this new version of the patch?

Not working on my end, apparently. Exact same behavior as discussed in second half of #643.
## Screenshots
screenshot_2018-08-19 gargoyle router management utility
screenshot_2018-08-19 gargoyle router management utility 1

## Latest build
screenshot_2018-08-19 gargoyle router management utility 2

lantis1008:
The build you are running is before the patch was applied to source. 
I’ll take a look but I think maybe it is ok.

d3fz:
>The build you are running is before the patch was applied to source.

Thanks for clarifying that. Consider it solved, then.

lantis1008:
I’ll still check it. 
Before th patch was applied there should have been zero ipkg cleaning occurring. So it should have just worked. 

Spooky lol

lantis1008:
I just checked. It looks good on my end.
I will note that it is only being applied to ar71xx target currently. 

I think you should try a later build and let me know. But for now i'm going to write this one off.

d3fz:
Alright, I'll wait for the next build and report back.

Thanks for the feedback!

PR #656, opened 2017-09-21
Fix Spectrum Analyser plugin + Restore strings to firstboot

lantis1008:
Fix various calculation fails for spectrum analyser plugin and stop the table from escaping under mobile conditions.
Also restore the strings to the firstboot (non i18n) page. I think these went missing during the large UI update patches.


d3fz:
Please also fix spectrum_analyser.sh page.

Spectrum table is escaping it's parent when going mobile (e.g. 360px).

This should fix the problem:
`<div id="spectrum_table_container" class="table-responsive"></div>`

lantis1008:
done, thanks @d3fz

PR #682, opened 2017-11-14
Update basic.js

fengchen-github:
fix x86_with_no_wireless return mac80211Channels undefined


lantis1008:
I don't think this is the full solution. If a device has no wireless then we should be removing all wireless options entirely.

lantis1008:
Hi, can you please test #692 and if satisfied, close this PR?
Feedback welcome and i appreciate your assistance in testing.

fengchen-github:
Ok, I'm testing

PR #687, opened 2017-11-18
Suggestion for code style fix.

fmmedeiros:
Make it explicitly that the else statement does not belong to the first if statement starting at line 892.


lantis1008:
Thanks for the changes. 
httpd_gargoyle is deprecated as of Gargoyle 1.7.2, this file is no longer used.

fmmedeiros:
Ok, thanks.

PR #693, opened 2017-11-28
add x86/x86-64 target

fengchen-github:


lantis1008:
My only comment is it might be better to make an x86 target, with default profile, and 86-64 be another profile underneath it. 
Then we get both. 

Only a suggestion though.

lantis1008:
Oh and you need to provide an “images” file as per other targets

obsy:
Please include by default kmod-usb-hid - required for usb keyboard.

IMO albo should be include - kmod-usb2/kmod-usb2-pci, kmod-usb3. Space is not problem so you can also include kmod-ath10k with firmware

fengchen-github:
There is a target alix is x86, but I do not know what it is  @lantis1008

fengchen-github:
ok, @obsy

ispyisail:
@ericpaulbishop what happening with this patch?

I want to try it out

ericpaulbishop:
I may add this when after updating the new version of Openwrt but adding before then means supporting it in both the old Openwrt version and the new. So, this is on hold until the update is done (which I'm working on now)

lantis1008:
One issue is that packages don't get grabbed correctly, and i haven't thought of an elegant solution yet.

in `*-src/bin/packages` there are 2 folders rather than the typical 1. This breaks the `find` logic in part of `build.sh`.
One possible solution would be to recursively copy all package files across, and concatenate the Packages manifest etc files together. I am unsure how the signature would be treated in this case

I've created a branch for rebasing this to 1.11.x (https://github.com/lantis1008/gargoyle/commit/a1602a98bb95847ace79b91db4225b7de3ccc9bf) but i haven't been able to run test it, only compile test. I also haven't included any of the patches or additional requests from this thread at this stage. Hopefully it is a good reference for the future changes.

lantis1008:
Please close in favour of #774 
If you need to make any changes please base them on the new developments.

Thanks.

PR #724, opened 2018-02-02
Fix popup window centering on edit pages

d3fz:
An attempt to fix edit popup pages centering by making use of the `window` properties described below and some minor code changes: 

>Window.outerWidth/outerHeight gets the width/height of the outside of the browser window. It represents the width/height of the whole browser window including sidebar (if expanded), window chrome and window resizing borders/handles.

Browser compatibility should not be a problem. Although, Microsoft Edge seems to do it's own thing when it comes to window positioning, which is a known issue apparently. Works fine with every other browser I tested it with (Chromium based, Firefox, etc).

## Screenshots (Before/After)
offcenter
centered
offcenter2
centered2

### DIFF
--- a/package/gargoyle/files/www/js/bandwidth.js
+++ b/package/gargoyle/files/www/js/bandwidth.js
@@ -864,17 +864,7 @@ function expand(name)
 		expWindow = null;
 	}
 
-	try
-	{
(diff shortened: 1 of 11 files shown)


obsy:
+1

PR #725, opened 2018-02-09
Add basic HTTPS sniffing capability to Gargoyle

lantis1008:
- Make Gargoyle Webmon (Status->Web Usage) more useful by allowing it to see the destination URL (FQDN only) of **most HTTPS requests.

- Make Gargoyle Weburl more useful by allowing it to match the destination URL (FQDN only) of **most HTTPS requests. Can now place restrictions on websites with reasonable levels of success.

- Increase flexibility of make_iptables_rules (and fix a bug i introduced 2 months ago unknowingly)
- Fix a missing reject line for URL matching

- [Add in ](https://www.timeanddate.com/time/zone/turkey)[Turkey Time](https://i.ytimg.com/vi/TQOJmOX3FYk/maxresdefault.jpg) [as per change in the local timezone (2016, we missed it by a bit)](https://www.timeanddate.com/time/zone/turkey)

** This is achieved by leveraging the Server Name Indication (SNI) extension of the TLS1.2 standard in HTTPS authentication. The host name (e.g. example.com) is served in the clear in the packet so that the server knows which certificate to send back to the client. The path (e.g. /foo.html) is only sent once the transmission is fully encrypted. That is to say, we can only match the domain of HTTPS traffic.
SNI is supported by all major browsers and has been gaining proliferation since the mid 2000's.


lantis1008:
@ericpaulbishop 
Am i crazy, or is there a missing REJECT statement here matching against 0xFF000000/0xFF000000: https://github.com/ericpaulbishop/gargoyle/blob/master/package/gargoyle-firewall-util/src/make_iptables_rules.c#L547

I think this is the missing piece of the puzzle to getting this module finished. I'm only just beginning to understand how all of the Gargoyle chains run together (thank you for your diagram/explanation).

lantis1008:
@ericpaulbishop This is now ready for merge.
As to my previous comment, after testing i have concluded i am correct. Or at least, the behaviour is correct after adding in the proposed fix.

ericpaulbishop:
Matching the domain on the (cleartext) SSL certificate is really clever -- I like that a lot. My thought is that we would handle the problem by matching DNS requests (and we may still add that at some point to block other types of traffic), but this is better in that it preserves the exact same functionality as before.

PR #729, opened 2018-02-16
Visual update - add headings

WildRat:
Made visual style more unified


lantis1008:
Can you seperate this at all? It also has the quad9 change

Issue #654, opened 2017-09-09
Themes broken by new GUI


lantis1008:
Need to bring themes inline with new UI
- [ ] Gargoyle Classic
- [ ] Gargoyle Modern (half done, could use polishing)
- [ ] Green Theme
- [ ] Light Theme
- [ ] Theme by Matei
- [ ] Dark One
- [ ] Demantoid
- [ ] Horchata
- [ ] Xeye

lantis1008:
Provided basic screenshots to assist

d3fz:
How's the progress so far ? I'm having a go with "Xeye" (halfway there) and "Theme by Matei", based on your changes to Light/Green Themes. 

![screencapture-192-168-1-1-access-sh-1505179939691](https://user-images.githubusercontent.com/22578839/30304008-a0b46aec-9741-11e7-9485-52ecb47d248c.png)


Not sure how to approach "Demantoid"/"Horchata" due to custom menu layout.

lantis1008:
Thanks for doing that.
I'm also avoiding those two for that exact reason, but I will get to them eventually. 
I'm having a look at cleaning Gargoyle-Modern as it got kind of forgotten in the process of getting Gargoyle-Classic up to spec.

d3fz:
No problem.
Xeye is pretty much finished. 
Having a look at the other one now.

d3fz:
By default, these themes were all centered with width ~ 900px, "wasting" useful available space. 
Should we make then full-width or leave as it is ?  

Just need to make sure it doesn't break when going mobile, and some minor tweaks.

After that is pretty much finished.

lantis1008:
Agreed. I made them 1300px wide (if I remember correctly) to give the same *feel* as the old themes, whilst maximising space. 
This size auto shrinks as the window does, so no usability issues with the themes I've done. 

up to you, if you want a new spin on it which is full width and not centered, go for it.

d3fz:
Yes, increasing max-width to 1300px makes it a lot better, so i'm just gonna leave as it is. I actually like the centered layout.

Once i'm finished, where/how can i send it to you ?

lantis1008:
You can create your own PR if you like, or send via a file sharing site etc. no problem

d3fz:
I've tried to match them as close as possible without breaking anything, looking pretty close. 

There you go: https://ufile.io/726bh

I'm sending via shared file so you can take a look and perhaps include into your already opened PR.

[...]

PR #734, opened 2018-02-25
Allow bandwidth graph colours to be set by theme

lantis1008:
Pulls the "color" attribute of the label for each plot and applies it to the polyline object color and fill attributes


d3fz:
Nice. 👍 

Can you provide any screenshots showing how it looks? 

For some reason, it's not working on my end. When setting a different color, only the labels color change. This should also affect graphs color, right?

Clearing browser cache doesn't seem to fix it.

lantis1008:
bccb7114-7803-413c-80f4-e29ee324d943

You took the changes from svg_bandwidth.js as well?
This is a quick change with softer colours

d3fz:
>This is a quick change with softer colours

Those graphs colors look smooth on a dark background. Nice.

>You took the changes from svg_bandwidth.js as well?

Yes. 

After further testing, console was throwing `TypeError: window.parent.getElementById is not a function`

Changing to `window.parent.document.getElementById` fixed the issue.

___

Oddly enough, `bandwidth_expand.sh` header has some duplicate code:

```Javascript
gargoyle_header_footer -m -c "internal.css" -j gargoyle_header_footer -m -c "internal.css"
````

lantis1008:
Thanks I’ll double check that error. I wasn’t seeing it. 
What browser?

d3fz:
Vivaldi (Chromium based).

Also tested with latest Chrome/Opera/Edge and Firefox Quantum. Same problem.

lantis1008:
Thanks for pointing it out. Was a typo in the commit, was fine on my router (hence why i saw no bugs). Whoops.

ericpaulbishop:
Looks great, merged.

lantis1008:
Whoops. Accidentally pushed the theme as well. 
I’ll do a commit to enable it for building later this week.

d3fz:
I was about to make a comment on that, nice work on the horizontal menu bar by the way. Maybe now it should be possible/easier to address those two remaining themes in #654?

lantis1008:
Definitely. They are next on my todo list :)

PR #744, opened 2018-03-19
Fix image copy looking in wrong directory + others

lantis1008:
Think we need to look in `bin/targets` rather than `bin`.
```bash
$ ls bin
packages\ntargets
```

`find: 'bin/targets/packages\ntargets/': No such file or directory`

Think this still works with multiple profiles being built (have not tried yet)

Fixes 049465778bfe0c88e49665c690f666a28dc224b8 and 263ec07987a05720a857e3238a88f64ce3f9e1d7

---------------------------------
Bump the build number, and fix switchinfo and make it target generic (still need to define the board names for the port numbers)


ericpaulbishop:
Thanks! I tested the code snippet but with hard coded arch without doing an actual build. Thanks for catching that

PR #745, opened 2018-03-20
Fix page loading for several pages

lantis1008:
gargoyle_header_footer -i switch now seems to be location sensitive. Ensure that the list of UCI config files to load is listed last.
Perhaps this was always wrong, and some compiler magic let it slide.


ericpaulbishop:
Thanks. My guess is that this has to do with the fact that the getopt() function used to parse options is defined in the standard C library, and the newest version of OpenWRT uses musl instead of uclibc as its standard C library implementation.

PR #746, opened 2018-03-22
Rewrite Switchinfo + Openvpn Updates

lantis1008:
Rewrite of Switchinfo.sh to be device independent
- Should now support any device with a valid /etc/board.json file with a mapping between the logical and physical ports
- Can support > 4 LAN ports (although >9 may require more cosmetic testing, cross that bridge later)
- Admittedly have only tested on a few devices, but don't foresee any major hassles

Openvpn Updates
- Remove BlowFish ciphers. These have been found to be insecure (see SWEET32 Birthday attacks)
- Remove all references to keysize parameter (was only required with ciphers like BF)
- Set default Cipher to AES-256-CBC (should be used for backwards compatibility concerns with older clients). There is an option to allow outdated clients to negotiate the use of BF ciphers, however i have chosen not to allow this as it is to be completely removed in following updates of OpenVPN
- Add new ciphers AES-128/256-GCM (note that newer clients will negotiate with server and upgrade cipher to AES-256-GCM if capable on both ends)

### DIFF
--- a/package/gargoyle/files/usr/lib/gargoyle/switchinfo.sh
+++ b/package/gargoyle/files/usr/lib/gargoyle/switchinfo.sh
@@ -2,94 +2,63 @@
 
 #
 # (c) 2013 Cezary Jackiewicz, http://eko.one.pl
+# (c) 2018 Michael Gray
 #
 
 [ -e /sbin/swconfig ] || exit 0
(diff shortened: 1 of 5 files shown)


obsy:
About switchinfo: tested on several devices and looks like it works.

PR #752, opened 2018-04-12
Wireless Regulatory Settings (revisited)

lantis1008:
This is a revisit of #469 (now closed) which was initially rejected, and then reopened on condition of changes requested.  

> Allows the user to select the country they are in for regulatory purposes. The default option ensures regulatory compliance in all countries

Through the use of a new tool Gargoyle-IP-Query (leveraged off existing code from ddns_updater), we can attempt to detect the GeoIP location of the user.

The user will be allowed to set a wireless country setting under the following conditions:
- A GeoIP location has been successfully detected (requires internet connection)
  - For this reason, this option is not available at firstboot, but will show up the first time Connection -> Basic is loaded after an internet connection is available
- If the country setting has been previously set, it cannot be set again
  - This includes routers which now ship with the country setting by default

The user is also presented with a large text block:

> Gargoyle has attempted to determine your geographical region. If this is not correct, it is strongly recommended to leave this setting as World (Default).
> Setting this value incorrectly may lead to you violating regulatory standards in your region. If in doubt, choose World (Default) which is legal in all regions.
> By setting your country correctly, you may be allowed to use additional channels, higher transmit power or wider channel widths.
> 
> Note: You may only change this setting once.

Would appreciate testing here. I believe i have covered most bases.
If this still does not fit the brief, please let me know what you would like to see differently.


francisuk1989:
Great work and great idea about GeoIP location

ericpaulbishop:
Looks great, merging now.

PR #753, opened 2018-04-23
more mvebu work

lantis1008:


ericpaulbishop:
Your 4.14 imq patch doesn't handle the gargoyle specific parameters to determine where it gets hooked. 

I just committed a different version, that does handle this (but still needs testing).

I cherry-picked the other mvebu updates (thanks!), and am going to close this pull request.

lantis1008:
Ahh. And that is probably the source of my boot loop. Thanks mate

PR #761, opened 2018-07-14
Update targets. Fix zip utility. Bump to 18.06.0

lantis1008:
brcm47xx compile tested only. No runtime.


lantis1008:
not yet ready for merge.

lantis1008:
@ericpaulbishop 
Please see the following Google Sheet
https://docs.google.com/spreadsheets/d/1E44rmp-F6B5P0W6TQ1GNmVFxjCIYCv88BYAqyuSlAAY/edit?usp=sharing

Devices highlighted in Green have been successfully compiled under the current code base.
- I propose that devices labelled Tiny (4mb flash) be removed from the picture entirely.
- There's some reasonably popular routers in the USB profile which aren't compiling. I think the kernel is exceeding capacity (rather than the rootfs), as i tried removing openvpn as a default package from USB and got no further. Can either:
  - Turf these as well
  - Move them to default and see if they compile?
- Are there things that don't need to be in here that have been forgotten or aren't optimised correctly? I admit to not knowing too much in this area. I tried to pull IPV6 out but as you said, we have a few dependencies on it these days. It also seems to be much more tightly integrated into the Kernel.

lantis1008:
@ericpaulbishop 18.06.0 has been released and tagged.

PR #764, opened 2018-08-09
Update msmtp to 1.6.6. Fixes compile issue on some targets. Bump to 18.06.1

### LINKED ISSUE #766: Gargoyle m4 build fails with glibc 2.28
Gargoyle is pulling in openwrt 18.06 toolchain, and the m4 in that chain is broken in glib 2.2
Here is the m4 patch necessary to fix

https://src.fedoraproject.org/rpms/m4/raw/814d592134fad36df757f9a61422d164ea2c6c9b/f/m4-1.4.18-glibc-change-work-around.patch

lantis1008:
- Update msmtp to 1.6.6 to fix some compile issues
- Bump to 18.06.1 release (probably need to bump again to pull in dropbear CVE fix)
- Properly fix gpkg sources for Openwrt repos

Closes #766

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -33,7 +33,7 @@ set_version_variables()
 	# set precise commit in repo to use 
 	# you can set this to an alternate commit 
 	# or empty to checkout latest 
-	openwrt_commit="75e4d2d18c0c6a604cc9bf7bce2d6023fcedc78a"
+	openwrt_commit="70255e3d624cd393612069aae0a859d1acbbeeae"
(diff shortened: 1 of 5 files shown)


hungaborhorvath:
I can confirm that I managed to build a working image with this for youku-yk1 openwrt target profile. The msmtp problem was indeed fixed, for details see thread https://www.gargoyle-router.com/phpbb/viewtopic.php?f=8&t=11430

PR #774, opened 2018-09-08
Add x86 generic and 64bit targets for 1.11.x

lantis1008:
Rebase of #693 without the ath10k support (can be added) and no patches
Compile and runtime tested


lantis1008:
Still needs the packages retrieval script modified.
Current thoughts are to scrape ARCH_PACKAGES out of the .config file to figure out what folder to look in

PR #776, opened 2018-09-24
Fix node.js versin variable

mysticall:
I'm not sure, but I think the variable $node_tag should be replaced with $node_version_tag.


lantis1008:
Thanks, this is merged with 73cde5343cd2cd0661b806f417b4f39af524b3ae.
Please close @mysticall .

PR #779, opened 2018-10-03
Modified some configurations on x86 / x64

fengchen-github:
Tested on virtualbox and intel j1900


lantis1008:
I’m not opposed to the other changes if they are resubmitted with better formatting just so I’m clear. 
I don’t want to discourage your contribution and assistance.

PR #785, opened 2019-01-17
Update openvpn_download_credentials.sh

prezeskk:
Not adding key-direction to config file

### DIFF
--- a/package/plugin-gargoyle-openvpn/files/www/utility/openvpn_download_credentials.sh
+++ b/package/plugin-gargoyle-openvpn/files/www/utility/openvpn_download_credentials.sh
@@ -32,7 +32,7 @@ $(cat /etc/openvpn/client_conf/$GET_id/$GET_id.crt)
 $(echo -e '</cert>\n<key>')
 $(cat /etc/openvpn/client_conf/$GET_id/$GET_id.key)
 $(echo -e '</key>\n')
- $keystring
+$(echo ' '$keystring)
 $(echo -e '<tls-auth>')
 $(cat /etc/openvpn/client_conf/$GET_id/ta.key)
 $(echo -e '</tls-auth>')


prezeskk:
On TP-Link TL-WDR4300 not working. There is no space at the beginning line
and key-direction 1 is missing.

lantis1008:
Ok. I don’t have that device to test, so I will need to take your word for it. 
I will check your change and if they cause no regression in behaviour I’ll bring it in.

I’ll test and get back to you. Thanks

lantis1008:
Thanks, i have merged your change into the current working branch (base_on_openwrt_remerge) with this commit
https://github.com/ericpaulbishop/gargoyle/commit/35e53238bbfc579d49ae959ba39b5b7633ce56a8

You can either close this pull request, or let it auto close when base_on_openwrt_remerge is merged with master.

prezeskk:
thx

ispyisail:
Is this a good idea?

> From what I understood it is possible to omit the key direction, which will result in bidirectional use of keys: one HMAC and one for encryption and decryption. 

> 
Alternatively is it possible to configure altering values between 0 and 1 on the server and the client. (e.g. on the server 0 and on the client 1, or the other way around). Which causes the use of 4 distinct keys (HMAC-send, cipher-encrypt, HMAC-receive, cipher-decrypt).

lantis1008:
Unless the option has changed, you can set key-direction = 0,1, or bidirectional. This is what the script currently does. 
If bidirectional is no longer explicitly stated then another change is required to do that. 
I’d need to double check behaviour again but I think we already use the 4 way encryption.

lantis1008:
Just as a follow-up:
- OpenVPN changed the key-direction parameter so it is now part of tls-auth i.e. `tls-auth $file $keydirection`
  - If $keydirection is omitted, bidirectional is used
- Gargoyle sets the server to use key-direction 0 [here](https://github.com/ericpaulbishop/gargoyle/blob/base_on_openwrt_remerge/package/plugin-gargoyle-openvpn/files/usr/lib/gargoyle/openvpn.sh#L201-L229)
- Gargoyle sets the client to use key-direction 1 [here](https://github.com/ericpaulbishop/gargoyle/blob/base_on_openwrt_remerge/package/plugin-gargoyle-openvpn/files/usr/lib/gargoyle/openvpn.sh#L391-L408)
- When generating the single .ovpn file for clients, we interpret the Gargoyle settings and either set it as bidirectional, or as the corresponding 0 or 1 [here](https://github.com/ericpaulbishop/gargoyle/blob/base_on_openwrt_remerge/package/plugin-gargoyle-openvpn/files/www/utility/openvpn_download_credentials.sh#L19-L24)

So behaviour is solid in my opinion.
HOWEVER, i do not know if the `bidirectional` symbol is still valid, or whether you omit it entirely in this case. I'm seeing conflicting reports on this.
I implemented it based on the FAQ [here](https://openvpn.net/faq/i-am-having-trouble-importing-my-ovpn-file/)

PR #786, opened 2019-01-21
Add support for Xiaomi Mi Router 3G

naseef:


naseef:
Compiles ok. Flashed successfully.

obsy:
@lantis1008: little OT: maybe we should transfer all common options to another file, say, "common" - like CONFIG_CLEAN_IPKG_PARTIAL=y, CONFIG_BUILD_PATENTED=y, CONFIG_PACKAGE_plugin-gargoyle-usb-printer = m etc, and combine it into one .config before building the image for target.

lantis1008:
@obsy we could try using diffconfig rather than these full config styles? This would shrink the files significantly and make them easier to update

lantis1008:
I have merged this with additional changes to add ARPING utility.

PR #787, opened 2019-01-23
preserve timestamp while copying gargoyle packages

naseef:
This patch avoids rebuliding of copied gargoyle packages while rebulding and saves time [ PS: I'm new to this and correct me if I'm wrong ]

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -543,7 +543,7 @@ for target in $targets ; do
 			fi
 		done
 		IFS="$IFS_ORIG"
-		cp -r "$package_dir/$gp" "$target-src/package"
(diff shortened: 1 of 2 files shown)


lantis1008:
Although, any package with a JS or CSS file will be prone to at least 1 rebuild due to the compressed files generating with new timestamps. 
To make this complete, I suppose these could be faked, but for now it can be left as a limitation.

lantis1008:
Merged, thanks.

lantis1008:
Thanks very much for this change. I didn't realise how much time we were wasting on each build. This is significantly better.

PR #788, opened 2019-01-24
Enabled support for v3 of tl-842n router

teambob:


teambob:
Hi,

I have it up and running now on the usb profile. What is the difference with the usb_large profile?

lantis1008:
What level of free space do you see on the plugins page?

The usb_large profile has additional functionality built in that does not fit on small devices:
- QoS
- upnp
- wol
A few others. 

You can still use this functionality on yours, but would need to install it after flashing. 
So this is more a convenience change than anything.

teambob:
Root Drive 9.125 MBytes Total, 8.680 MBytes Free

Happy to move the changes to usb_large for the broader benefit

lantis1008:
If you could please take the time to move it to usb_large, recompile and check that it still is ok and the free space is sufficient, that would be great. 
If it is all good, please update the pull request and I’ll merge it. 

If the test is unsuccessful, let me know and I’ll merge it in the current form. 

Thanks!

teambob:
Hi,

I have updated and tested the new patch. I can see the new functionality :)

-Andrew

PR #791, opened 2019-02-09
Fix fetching bw data which requires multiple runs

lantis1008:
@obsy Can you please review this change? I'd normally just push it straight through but it's a bit of a hairy one.

**Symptom**: Users reported [here](https://www.gargoyle-router.com/phpbb/viewtopic.php?f=12&t=11764) that the High Resolution 15m graphs would only show some IP addresses after a period of time.

**Problem**: It appears that once the data in this monitor gets large enough that it can no longer be fetched in a single query, it is failing to do so.
The main cause of this is that the `buf` which we are sending back into the socket has now been filled with response data which is in a different format.

**Solution**: The quickest solution to this is to zero out the buffer and set it up again for the next request [Lines 449-463](https://github.com/lantis1008/gargoyle/blob/3b74ef1fb605e9b2c1de3e671a69649877604107/package/libiptbwctl/src/ipt_bwctl.c#L449-L463)

**Problem**: The next issue that becomes apparent is that we are expecting the request_index to be at `buf+4` [Line 382](https://github.com/lantis1008/gargoyle/blob/3b74ef1fb605e9b2c1de3e671a69649877604107/package/libiptbwctl/src/ipt_bwctl.c#L382), however in the response it is at `buf+5` [Line 1716](https://github.com/ericpaulbishop/gargoyle/blob/fdfcfe3e0dcb054266b3743b79607db703cc2b86/netfilter-match-modules/bandwidth/module/ipt_bandwidth.c#L1716).

**Solution**: The easiest solution to this is to keep track of the index in a separate variable [Line 445](https://github.com/lantis1008/gargoyle/blob/3b74ef1fb605e9b2c1de3e671a69649877604107/package/libiptbwctl/src/ipt_bwctl.c#L445)

**Problem**: With these two problems fixed, it then becomes apparent that there is a logic fail which causes the data to be overwritten each time because we never indicate that we have successfully initialized the data.

**Solution**: Toggle the flag [Line 436](https://github.com/lantis1008/gargoyle/blob/3b74ef1fb605e9b2c1de3e671a69649877604107/package/libiptbwctl/src/ipt_bwctl.c#L436)


obsy:
Looks probably good. Can this user test it before commit?

lantis1008:
Maybe. I’d probably need to supply a build for them. I assume libraries are built target specific so I can’t just send the library for mine.

PR #793, opened 2019-02-24
i18n/Polish: update translation

obsy:


lantis1008:
Merged, thanks.
I'm still looking into your other query about rebind lists

Issue #795, opened 2019-03-03
[RFC] Rename obfsproxy to obfsproxy-legacy


obsy:
Why? Because in openwrt repo (ie. src/gz openwrt_18.06-snapshot_packages http://downloads.openwrt.org/releases/18.06-SNAPSHOT/packages/mips_24kc/packages)  there is already a package called obfsproxy in a newer version and the plugin is trying to install a new version with a lot of dependencies (python ...)
```

root@Gargoyle:~# gpkg info obfsproxy
Package: obfsproxy
Version: 0.1.4-1
Install-Destination: Not Installed
SHA256sum: d3eda785d034ba2fdb07546feff2d1b34111aa275c54d60fe8f923525ac136e4
Size: 23147
Installed-Size: 22560
Description: Tor obfsproxy Plugin
Architecture: mips_24kc
Source-ID: gargoyle_base
Section: net
Filename: obfsproxy_0.1.4-1_mips_24kc.ipk
Status: unknown ok not-installed
Depends: libc, libevent2, libopenssl, libpthread, librt

Package: obfsproxy
Version: 0.2.13-1
Install-Destination: Not Installed
SHA256sum: dfc3b11f3a1ae1b3f50e7cb59ef766983e1184a4d169f3cd20f7fc322784421d
Size: 75068
Installed-Size: 74043
License: BSD-3-Clause
Description: obfsproxy is a tool that attempts to circumvent censorship, by
             transforming the Tor traffic between the client and the bridge. This
             way, censors, who usually monitor traffic between the client and the
             bridge, will see innocent-looking transformed traffic instead of the
             actual Tor traffic.
Architecture: mips_24kc
Source-ID: openwrt_packages
Section: net
Filename: obfsproxy_0.2.13-1_mips_24kc.ipk
Status: unknown ok not-installed
Depends: libc, python-light, python-crypto, python-pyptlib, python-setuptools, python-yaml, twisted

root@Gargoyle:~# 


root@Gargoyle:~# gpkg info -v 'Install-Destination,Required-Size,Required-Depends,Description,Will-Fit,User-Installed' -d plugin_root -o 'js' -r /^plugin\-gargoyle\-tor/
pkg_info = [];
pkg_info["plugin-gargoyle-tor"] = [];
pkg_info["plugin-gargoyle-tor"]["1.11.0-1"] = [];
pkg_info["plugin-gargoyle-tor"]["1.11.0-1"]["Install-Destination"] = "Not Installed";
[...]

PR #796, opened 2019-03-03
Rename obfsproxy to obfsproxy-legacy

obsy:
See https://github.com/ericpaulbishop/gargoyle/issues/795

### DIFF
--- a/package/gargoyle-tor/Makefile
+++ b/package/gargoyle-tor/Makefile
@@ -121,7 +121,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gargoyle-tor
 PKG_VERSION:=0.3.2.10
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=tor-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://www.torproject.org/dist \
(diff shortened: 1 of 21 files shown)


lantis1008:
Is this going to need updates to all configs as well?

obsy:
Yes, good point.

lantis1008:
We might need to pull the same trick with Openvpn. Looks like on new installs it is pull 2.4.5 from upstream, but we only have 2.4.4, and this ruins the init.d file.

obsy:
openvpn -> gargoyle-openvpn ?

lantis1008:
I think so

lantis1008:
I'm taking care of this now. Just realised i went with "openvpn-gargoyle" instead though. 
Change is difficult due to package variants.

obsy:
Gagroyle uses only openvpn-openssl. So we can rename openvpn to openvpn-gargoyle (or gargoyle-openvpn) and drop variants?
easyrsa also require changing name.

lantis1008:
Could do, but i'd like to leave the possibility open for future, so leaving variants in.
easy rsa seems to be ok because they have version "3", but we have version "2013-xx-xx" which is considered larger "newer" than 3.
But i would like to investigate using 3 anyway so i'll touch on that soon.

lantis1008:
would be great if you could take a quick look at https://github.com/ericpaulbishop/gargoyle/pull/807 before merge.
easy-rsa will be another day, as i don't think it causes an issue (yet).

PR #809, opened 2019-03-09
Typing errors and repairs.

RomanHK:
Typing errors and repairs.
I have allowed myself to fix the 1st error in the English version.


lantis1008:
Merged.
Is it possible in the future to squash all of these commits into a single change? Typically where the changes are all related we would have them as a single commit.

RomanHK:
Unfortunately, I can not join a single commitment and I can not synchronize your changes with my forked repository.

PR #820, opened 2019-06-01
Add OpenVPN to the backup

obsy:
OpenVPN / configs / clients certificates should also be added to the backup.


lantis1008:
Thanks, merged

PR #835, opened 2019-10-10
Increase QoS tc divisor to 512

Phyxion:
I'm having similar performance issues with QoS as described [here](https://www.gargoyle-router.com/phpbb/viewtopic.php?t=11961). Increasing the tc divisor increases my throughput by almost 10%. Further increasing it didn't do anything _for my connection_ (YMMV). This increase does seem to cost a bit of additional memory usage (<1MB, checked immediately after boot) but it should be OK as there's still 100MB of free memory left.

### DIFF
--- a/package/qos-gargoyle/files/qos_gargoyle.init
+++ b/package/qos-gargoyle/files/qos_gargoyle.init
@@ -455,7 +455,7 @@ initialize_qos()
 			#Add the leaf class
 			tc class add dev $qos_interface parent 1:1 classid 1:$next_class_index hfsc ls m2 ${m2}Mbit $ll_str $ul_str
 			#Add the qdisc to the leaf class, assuming average packet at 250 bytes.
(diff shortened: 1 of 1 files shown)


lantis1008:
Thanks. 
I'll need to check this first. In my experience it made no difference, however I was using a powerful device to begin with. 
If you could post some before and after speed tests to go along with it, and also what device you're using, that will also help me with what changes I can expect.

Phyxion:
Yes, of course. I'm actually using the same device: TP-LINK Archer C7.

Before:
img

After:
img

I'm not sure if it matters, but there are about 20 different devices connected to the router. For a couple of devices I have a QoS rule that limits them to 10% of the available throughput (up/down).

lantis1008:
I haven't forgotten about this. Will try to setup a test environment in the next week to check this.

PR #842, opened 2019-11-12
Fix custom ntp servers order not being respected

lantis1008:
While checking if the previous list value matched the new one, the variables were being mutated and the sort operation was applying to the original variables, resulting in ordered lists every time.
This is not always helpful behaviour.
Change the behaviour to create unordered lists, and if we want them ordered it is up to us to order them before creating the list.

### DIFF
--- a/package/gargoyle/files/www/js/common.js
+++ b/package/gargoyle/files/www/js/common.js
@@ -548,12 +548,10 @@ function UCIContainer()
 				var matches = oldValue.length == newValue.length;
 				if(matches)
 				{
-					var sortedOld = oldValue.sort()
-					var sortedNew = newValue.sort()
-					var sortedIndex;
-					for(sortedIndex=0; matches && sortedIndex <sortedOld.length; sortedIndex++)
+					var matchIndex;
+					for(matchIndex=0; matches && matchIndex < oldValue.length; matchIndex++)
 					{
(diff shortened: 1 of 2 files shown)


lantis1008:
@obsy Looking for your opinion and checking of this change to make sure it makes sense and does not create unintended other consequences please.

This is in reference to the issue reported here: https://www.gargoyle-router.com/phpbb/viewtopic.php?f=6&t=12166

obsy:
Is the order in ntpd important? Based on this: https://git.busybox.net/busybox/plain/networking/ntpd.c all host are used to choose the best one.

lantis1008:
No it's not. 
From a GUI point of view it is surprising to the user, but no change in functionality. 

I thought this might also affect DNS, but we dont use lists there.

obsy:
OT: Resolver in musl also do it asynchronously https://wiki.musl-libc.org/functional-differences-from-glibc.html#Name-Resolver/DNS

lantis1008:
So I guess the question is whether we want to keep ordering list options or allow them to be any order?

obsy:
IMO if it does not matter, just allow to add/save in any order. Uci list is also unordered, without any guarantee of list order.

lantis1008:
Thanks. Merged. Will monitor for regressions but i think this one will be ok.

PR #848, opened 2020-02-10
Support fast roaming.

n3vu0r:
Hello,

I've added GUI support for OpenWrt/hostapd's fast roaming functionality. This feature was already requested in the [forum](https://www.gargoyle-router.com/phpbb/viewtopic.php?t=11708). In contrast to LuCI's fast roaming [setup](https://forum.openwrt.org/t/802-11r-fast-roaming-in-luci/11730), I tried to keep Gargoyle's GUI very simple, see below. This is possible since OpenWrt simplified a lot like deriving the mobility domain from the SSID or deriving a PMK locally when using a PSK or auto-discovering BSSIDs when using RADIUS in order to exchange a PMK derived from a dedicated pre-shared roaming key.

**As far as I understand** that means, when a Pre-Shared Key (PSK) is present, a Pairwise Master Key (PMK) can be derived locally without any inter-AP communication and thus no further setup is needed. When no PSK is used (RADIUS), a PMK will be derived from a pre-shared roaming key and exchanged between the APs. This usually requires to list the BSSIDs and NASIDs of all APs which makes the setup complex and error-prone. But OpenWrt/hostapd can also auto-discover these parameters by using wildcard/broadcast BSSIDs which requires us to just specify a pre-shared roaming key. **In short**, when a PSK is present, fast roaming can just be enabled. When no PSK is present, a pre-shared roaming key must be specified and the GUI's Javascript will take care of assigning the pre-shared roaming key to the `r0kh`/`r1kh` option lists.

Initially, only a single field is added to the wireless section at `gargoyle.lan/basic.sh` to enable fast roaming but only for `WPA(2) PSK` or `WPA(2) RADIUS` encryption. The same applies to the guest network except that no `WPA(2) RADIUS` is offered.

Encryption: None -> Fast Roaming: Not visible
01-none
Encryption: WPA2 PSK -> Fast Roaming: Disabled (default)
02-psk-disabled
Encryption: WPA2 PSK -> Fast Roaming: Enabled
03-psk-enabled
Encryption: WPA2 RADIUS -> Fast Roaming: Disabled (default)
04-eap-disabled
Encryption: WPA2 RADIUS -> Fast Roaming: Enabled -> Roaming Key: Empty
05-eap-enabled
Encryption: WPA2 RADIUS -> Fast Roaming: Enabled -> Roaming Key: Random
06-eap-random

The OpenWrt [documentation](https://openwrt.org/docs/guide-user/network/wifi/basic#fast_bss_transition_options) states that a roaming key is 128-bit. But since version 2.5 (I think), hostapd uses 256-bit keys, and for backward-compatibility it would derive less secure 256-bit keys from 128-bit keys. So the GUI requires 64 instead of 32 hexadecimal digits.

**Open questions**:

  * The fast roaming GUI setup works and I can verify with WiFiAnalyzer (Android) that it is enabled. I see `WPA2-PSK+FT` and `WPA2-EAP+FT` but I haven't yet tested if it actually works. The log of hostapd should avoid the 4-way handshakes when fast roaming works. **Help needed.**
  * OpenWrt defaults to `ft_oder_ds = 1`. Does this mean FT-over-Air is disabled or are both enabled? Are there reasons to not enable `ft_over_ds`?
  * OpenWrt defaults to `pmk_r1_push = 0`. I enabled it when using RADIUS. It's supposed to speed things up. Are there drawbacks?
  * I've added and translated three new language variables, `FstTr`, `FtKey`, and `HexCh`. I think they are fine, but I'm speaking only German and English.
  * Currently hostapd 2.7 is packed with OpenWrt/Gargoyle. Version 2.9 has some fast roaming [improvements](https://w1.fi/cgit/hostap/plain/hostapd/ChangeLog).


lantis1008:
Thanks for this. I'm happy with the code, but lets discuss some of your questions first before we go ahead and merge this.

> OpenWrt defaults to ft_oder_ds = 1. Does this mean FT-over-Air is disabled or are both enabled? Are there reasons to not enable ft_over_ds?

1 enables it, 0 disables it. I don't see any reason to disable it.

> OpenWrt defaults to pmk_r1_push = 0. I enabled it when using RADIUS. It's supposed to speed things up. Are there drawbacks?

I don't know the history behind this mechanism. It defaults to off, so i'd like to see a good reason to enable it. Anecdotal is ok (within reason).

--------------
Also, when ready i'll be merging this against 1.12 rather than master. I'll also then forward port it to 1.13. For extra credit, i would greatly appreciate if you let me know whether in the WPA3 case i should treat it the same as WPA2 or not?

n3vu0r:
> > OpenWrt defaults to ft_oder_ds = 1. Does this mean FT-over-Air is disabled or are both enabled? Are there reasons to not enable ft_over_ds?
> 
> 1 enables it, 0 disables it. I don't see any reason to disable it.

I've just found an old latency [measurement](http://www.mikealbano.com/2014/06/80211r-80211k-fast-bss-transition.html):
>  non-802-11r: 150ms
802.11r-Over-the-Air: 16ms
802.11r-Over-the-DS: 71ms

This might not be representative but maybe we should consider to offer it as:

`Fast Roaming: [Over Air|Over DS|Disabled]`

Otherwise, I'm fine with the default.

> > OpenWrt defaults to pmk_r1_push = 0. I enabled it when using RADIUS. It's supposed to speed things up. Are there drawbacks?
> 
> I don't know the history behind this mechanism. It defaults to off, so i'd like to see a good reason to enable it. Anecdotal is ok (within reason).

I've just noticed that almost everyone is enabling it when searching for some information, so I enabled it as well. But I'm fine with the default here, too.

> Also, when ready i'll be merging this against 1.12 rather than master. I'll also then forward port it to 1.13. For extra credit, i would greatly appreciate if you let me know whether in the WPA3 case i should treat it the same as WPA2 or not?

According to this [article](https://wlan1nde.wordpress.com/2018/09/14/wpa3-improving-your-wlan-security/), Fast Roaming with WPA3 is not (yet) supported:
> “Fast Roaming” (802.11r) was left out of the current WPA3-“Standard” (for personal and enterprise networks).

On the other hand the new [hostapd.sh](https://github.com/openwrt/openwrt/blob/openwrt-19.07/package/network/services/hostapd/files/hostapd.sh#L39) seems to support something as `FT-SAE`. But unless someone has some insights, I would only offer it for WPA2.

lantis1008:
Thanks for switching it to 1.12.
Can you please resolve the merge conflict, and then i'll bring it in.

lantis1008:
Thank you. I've also forward ported both of your changes to 1.13

n3vu0r:
Great, thank you.

MrHaroldA:
Great work on getting this merged! I know I shouldn't be asking this, but is this to be included in a newer 1.12.x build, or do we have to wait until 1.13 or even 1.14 to use this?

lantis1008:
This is included in the 1.12 branch, so an individual can build their own 1.12.x and enjoy this change.
@ispyisail is also still providing 1.12.x builds on the forums so you can try it out there as well.

At this stage I don't have a planned 1.12.1 release, i'll need to organise this with Eric.
It will be of course included in any future 1.13 or 1.14 (or whatever) release that comes out. Again, you can build your own 1.13.x today off the master branch as well.

MrHaroldA:
Tnx, found it! https://www.gargoyle-router.com/phpbb/viewtopic.php?f=14&t=12339&p=55966#p55966

I'll install it on my three Archers (WDR4300v1, C5v1, C7v2) somewhere in the next few days and test the fast roaming option. My phone tries to stay connected to a router way too long while one with way better strength is available nearby. It even starts using mobile data before hopping over.

PR #849, opened 2020-02-14
Fix password validation and revelation.

n3vu0r:
The Wi-Fi password field is used for both encryption methods PSK(2) and WPA(2) RADIUS. Currently only its length is checked (`8 <= len <= 999`). This is fine for RADIUS but not for PSK. Using a 64-character password with non-hex characters will cause hostapd to fail and the webpage to hang.

Also a `reveal` checkbox can be inverted by checking it and then reloading the page. Due to browser's auto-completion, the checkbox will be checked again but the `togglePass` function is not aware of this, inverting the behavior of the checkbox. 

Changes:

  * When PSK is selected, validate against `8 <= len <= 63 || (len == 64 && is_hex)`.
  * Revalidate when encryption method is changed (RADIUS secret can be become PSK).
  * Add `autocomplete="off"` to `reveal` checkboxes.


lantis1008:
Thanks

PR #850, opened 2020-02-16
i18n/Polish: translation update

obsy:
For master/1.13.x


lantis1008:
I saw on your forum the comment about larger images. Are you concerned that there are changes (Gargoyle specific) leading to the increase? A lot of it will be due to change in base kernel + userspace from openwrt.
I did include wpad-openssl in most targets, but we already had libopenssl in each target so it shouldn't contribute a large change.

Open to ideas/feedback.

obsy:
I still wonder what to do. First thought - build a system with minimal requirements, all other things (themes, adbock, qos, openvpn etc) users should install themselves. Second - build without qos and openvpn for 8 MB flash and with qos and openvpn for large flash.

lantis1008:
When i removed qos last time, users were upset. 
Openvpn doesn't add much as we already have libopenssl.

It would be better to move to mbedtls, the only thing we rely on that doesn't work with that is easy-rsa i think.

PR #852, opened 2020-02-19
Improve autocomplete and proofread.

n3vu0r:
While browsers ignore `autocomplete="off"` for both login and non-login password fields, there is a [new way](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#The_autocomplete_attribute_and_login_fields) to prevent non-login password fields from being autofilled with the currently saved password by marking them as those via `autocomplete="current-password"` for login  and `autocomplete="new-password"` for non-login fields. Browsers will suggest a randomly generated password for non-login fields instead of autofilling the saved one.

Currently, input validation is triggered `onkeyup` instead of `oninput`. This was necessary in the past due to inconsistencies across browsers but as part of HTML5 the behavior of `oninput` was standardized and is now [widely supported](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/oninput#Browser_compatibility). `oninput` additionally fires on cut/paste/autofill and is more responsive, try this [demo](https://mathiasbynens.be/demo/input).

Changes:

  * Append `autocomplete="current-password"` to login password fields (`login.sh`, `password_confirm_template `). **Updated**.
  * Append `autocomplete="new-password"` to non-login password fields.
  * Replace `onkeyup` with `oninput`.

### DIFF
--- a/package/gargoyle/files/www/templates/password_confirm_template
+++ b/package/gargoyle/files/www/templates/password_confirm_template
@@ -11,7 +11,7 @@
 
 				<div class="row form-group">
 					<span class="col-xs-12">
-						<input type="password" id="password" class="form-control"/>
+						<input type="password" id="password" class="form-control" autocomplete="new-password"/>
 					</span>
 				</div>
 			</div>
--- a/package/gargoyle/files/www/templates/password_confirm_template
+++ b/package/gargoyle/files/www/templates/password_confirm_template
@@ -11,7 +11,7 @@
 
 				<div class="row form-group">
 					<span class="col-xs-12">
-						<input type="password" id="password" class="form-control" autocomplete="new-password"/>
+						<input type="password" id="password" class="form-control" autocomplete="current-password"/>
 					</span>
 				</div>
 			</div>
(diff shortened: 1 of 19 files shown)


package/gargoyle/files/www/templates/password_confirm_template
```diff
@@ -11,7 +11,7 @@
 
 				<div class="row form-group">
 					<span class="col-xs-12">
-						<input type="password" id="password" class="form-control"/>
+						<input type="password" id="password" class="form-control" autocomplete="new-password"/>
```
lantis1008: This isn't correct.
Currently the only (?) use of this modal is to confirm that you wish to format an external USB drive. Therefore it requires your current password, not a suggested new one.

Since we may use this modal in different ways in the future, we may need to be careful here. But certainly at this point, it should be "current-password"

n3vu0r: Thanks for pointing it out. I will change it to "current-password".

n3vu0r: If the modal will be used differently in the future, it could conditionally include one of two more general templates, `current_password_template` or `new_password_template`, just for the input field or container.

PR #854, opened 2020-02-22
Fix Router Access page

n3vu0r:
* Update UCI after saving changes w/o redirect

    In contrast to `basic.js`, `access.js` misses to update `uciOriginal` after saving changes without redirect. Resetting data afterwards results in disabling all `remote_accept`s (Remote Web Access and Remote SSH Access) because when saving changes, the old firewall sections are deleted from `uciOriginal` before it is cloned into `uci` to which the new firewall sections are added. But in the end, `uciOriginal` must be updated to `uci` to reflect all the changes including the new firewall sections.

Found some other issues, more commits coming.


package/gargoyle/files/www/access.sh
```diff
@@ -129,6 +156,22 @@
 						<label id="disable_web_password_label" for="disable_web_password"><%~ DisablePassword %> <em>(<%~ warning %>)</em></label>
 					</span>
 				</div>
+
+				<div id="lockout_template" class="row form-group" style="display: none">
```
lantis1008: I don't understand why we have this template and then other sections that display the message? Can you please clarify how this works? I only spotted this based on the JavaScript which suspiciously transfers the inner HTML one one element to another. 
If this is indeed what it is doing, please consider loading the text directly from JS to save characters. Translated text is still available in that space.

package/gargoyle/files/www/js/access.js
```diff
@@ -12,26 +12,22 @@ var stopRedirect = false;
 
 function saveChanges()
 {
-	errorList = proofreadAll();
+	var errorList = proofreadAll();
```
lantis1008: You have added "var" in several places where it is not explicitly necessary. Are you doing this intentionally or do you have an IDE doing things automatically?
These take up 4 bytes each in flash which is at a premium. Consider removal if not explicitly needed.

package/gargoyle/files/www/js/access.js
```diff
@@ -59,21 +56,41 @@ function saveChanges()
...
-			addAccept( document.getElementById("local_https_port").value, document.getElementById("remote_https_port").value);
+			addAccept(localHttpsPort, remoteHttpsPort);
 		}
-		if(document.getElementById("remote_http_port_container").style.display != "none")
+		else
 		{
-			addAccept( document.getElementById("local_http_port").value, document.getElementById("remote_http_port").value);
+			remoteHttpsPort = "";
 		}
+		if(oldRemoteHttpsPort != remoteHttpsPort) restartFirewall = true;
```
lantis1008: One line IF statements still require brackets as per the coding guidelines for Gargoyle. Please fix, or consider refactoring into the accepted `x = y == z ? A : B;`

package/gargoyle/files/www/js/access.js
```diff
@@ -190,50 +253,27 @@ function saveChanges()
 		commands += restartUhttpd ? "killall uhttpd\n/etc/init.d/uhttpd restart\n" : "";
 		//document.getElementById("output").value = commands;
 
-		stopRedirect = false;
-		var param = getParameterDefinition("commands", commands)  + "&" + getParameterDefinition("hash", document.cookie.replace(/^.*hash=/,"").replace(/[\t ;]+.*$/, ""));
+		uciOriginal = uci.clone();
```
lantis1008: It concerns me that we are resetting before actually executing the changes. There could still be a fail in the Ajax call yet. Can you justify this one or revert to original call order?

package/gargoyle/files/www/js/common.js
```diff
@@ -1035,7 +1071,32 @@ function setElementEnabled(element, enabled, defaultValue)
 	}
 }
 
+function setElementReadOnly(element, readOnly, associate = null)
```
lantis1008: This has been previously handled just using the disabled function. What are the benefits? If this is better, should it be more generic to allow the bootstrap styles and or themes to control the look?

lantis1008:
Also your other code changes will go in this weekend, have not forgotten.

package/gargoyle/files/www/js/access.js
n3vu0r: Yes, fixed using brackets.

package/gargoyle/files/www/js/access.js
n3vu0r: I reverted to original call order. Additionally, I used another uciCompare clone from which the old firewall accept sections are removed in order to derive the script commands. In this way, uciOriginal stays original until changes are saved.

But, when I'm not wrong, it makes no difference at the moment. Ajax can indeed fail, but we don't check for success or failure. Every page checks for `req.readyState == 4` but this just means the request is done. `req.state == 200` tells you that the request itself was successful (not the commands which got executed). When I open a page, unplug the Ethernet cable, and then click Save Changes, I get `req.readyState == 4 && req.state != 200`. I think all pages should check for `200`. But for this page, there might be a race condition between the request and restarting uhttpd.

package/gargoyle/files/www/js/common.js
n3vu0r: This is used when the HTTP -> HTTPS redirect is selected to signify that the remote HTTPS is enabled but cannot be freely chosen. Instead, the remote HTTPS port must equal the local one since uhttpd is only aware of local ports and will redirect to the local HTTPS port even if the request came from a remote HTTP port.

Yes, it should be more generic. Instead of using dashes, I would prefer Bootstrap's [Readonly plain text](https://getbootstrap.com/docs/4.0/components/forms/#readonly-plain-text) input style as default and themes should be able to overwrite it. But it requires Bootstrap version 4.

package/gargoyle/files/www/access.sh
n3vu0r: The warnings can appear at three different places which caused it. Below the local web protocol, below the remote web control, and below the RFC 1918 checkbox. In case of the local web protocol, the reverse proxy help inside the warning template is displayed as well. I will move it into JS.

package/gargoyle/files/www/js/access.js
n3vu0r: Interesting, haven't thought of saving space here. It's just vim and was intentionally. I will remove it where possible.

package/gargoyle/files/www/access.sh
n3vu0r: Done.

package/gargoyle/files/www/js/common.js
n3vu0r: Moved the readonly style from JS to CSS. Now it can be overwritten by themes. Dashing is maybe not so bad, it seems "half" an input, still editable by its associated field.

package/gargoyle/files/www/js/access.js
n3vu0r: Adding error handling for failing Ajax (if at all) affects all pages and can be its own PR. Currently success looks the same like failure, the "Please wait.." dialog just disappears. At least, uciOriginal stays unchanged now and clicking Save Changes again is a valid retry.

package/gargoyle/files/www/js/access.js
DoubleOhmSeven: Just nit picking here but if those 4 bytes are a problem then the js files would be better off being properly minified; common.js script for example could save ~30% on the file size and access.js around 25%.

lantis1008: > Just nit picking here but if those 4 bytes are a problem then the js files would be better off being properly minified; common.js script for example could save ~30% on the file size and access.js around 25%.

They do get minified during compilation. The 4 bytes do not get factored out as the minifier does not have enough context to factor them out.

package/gargoyle/files/www/js/access.js
lantis1008: > At least, uciOriginal stays unchanged now and clicking Save Changes again is a valid retry.

This is the important part.

lantis1008:
Thanks.

Just a reminder, preferably new changes go into master, as that is where the future of Gargoyle lies. 
If you would like the change backported to 1.12 as well, please let me know in the pull request and i'll organise it. It is easier for me to go backwards rather than forwards in these cases.

n3vu0r:
Good to know, thanks!

package/gargoyle/files/www/js/access.js
n3vu0r: I played a bit with uglifyjs options. I think @DoubleOhmSeven meant we can do better than now by enabling `-c -m` flags.


|file              | now  |  -c -m |   %   |
|-------------  | -------  |-------  |------|
|common.js |60 KiB | 42 KiB | 30  |
|access.js    |24 KiB | 18 KiB | 25  |

By doing so, I also observed that global variable names are not minified without the `--toplevel` option, making the file size even larger when removing `var`s. I want to try the `--name-cache` option as well.

DoubleOhmSeven: Correct, but the minifier used is pretty terrible, if you use UglifyJS 3 (you can check https://jscompress.com/ for difference) then you'll notice a huge difference. For example, common.js of 1.12.0: 31.89% compression, saving 18.54 kb.

n3vu0r: I've also just found [Terser](https://github.com/terser/terser):

> `uglify-es` is [no longer maintained](https://github.com/mishoo/UglifyJS2/issues/3156#issuecomment-392943058) and `uglify-js` does not support ES6+.

> **`terser`** is a fork of `uglify-es` that mostly retains API and CLI compatibility
with `uglify-es` and `uglify-js@3`.

lantis1008: If there is a better or more up to date uglifyjs/CSS that will give additional savings, im happy to accept a patch or suggestions. 
But in my brief reading, I would be quite sceptical of it operating in a non breaking way with function mangling and some global's (i18n vars) given the way the js is not internally referenced (i.e. the files are included by outputs of the header footer program). So it would need to be thoroughly tested and shown working.

n3vu0r: Okay, I will give it a try and check what impact mangling will have.

PR #860, opened 2020-03-04
Minify in parallel, compress/mangle JS ES6+.

n3vu0r:
Improve JS minification by enabling compression `-c` and mangling `-m`. This should save around 175 kB in total for the `ath79.ath10k-large` profile. I have know idea, is this something?

I used the opportunity to parallelize both JS and CSS minification via `xargs`, respecting `BUILD_THREADS`, and to switch from `uglify-js` to [`terser`](https://github.com/terser/terser):

>`uglify-es` is [no longer maintained](https://github.com/mishoo/UglifyJS2/issues/3156#issuecomment-392943058) and `uglify-js` does not support ES6+.
>
>**`terser`** is a fork of `uglify-es` that mostly retains API and CLI compatibility
with `uglify-es` and `uglify-js@3`.

According to the documentation, the options `-c -m` are safe. Top level (global scope) variable names and function names are not mangled.

With mangling enabled, global variables should be avoided whenever possible as these won't be mangled, assuming they consume more characters than prepending `var` which will also be grouped for multiple local variables whenever possible.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -140,59 +140,57 @@ EOF
 
 do_js_compress()
 {
-	echo "Compressing and mangling JavaScript files..."
-	terser_arg1="$1"
-	terser_arg2="$2"
-
-	rm -rf "$compress_js_dir"
-	mkdir "$compress_js_dir"
-	escaped_package_dir=$(echo "$top_dir/package-prepare/" | sed 's/\//\\\//g' | sed 's/\-/\\-/g' ) ;
-	terser_batch=""
-	for jsdir in $(find "${top_dir}/package-prepare" -path "*/www/js") ; do
-		pkg_rel_path=$(echo $jsdir | sed "s/$escaped_package_dir//g");
-		mkdir -p "$compress_js_dir/$pkg_rel_path"
-
-		for jsf in "$jsdir/"*.js ; do
-			compress_jsf=$compress_js_dir/$pkg_rel_path/$(basename "$jsf")
-			terser_batch="$terser_batch $jsf -o $compress_jsf -c -m"
-		done
-	done
 	num_terser_threads=1
 	if [ "$num_build_threads" != "unspecified" ]; then
 		num_terser_threads=$num_build_threads
 	fi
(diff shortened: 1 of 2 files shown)


lantis1008:
Thanks, can you provide (via Pastebin) a comparison of common.js using both methods please? Would like to understand the difference.

n3vu0r:
`uglifyjs common.js -o common.old.js`
`terser common.js -o common.new.js -c -m`

[common.js](https://pastebin.com/4UZzebru)
[common.old.js](https://pastebin.com/Sc38cW8q)
[common.new.js](https://pastebin.com/LgSzQQpT)

lantis1008:
Cool, so the big difference appears to be that variables internal to functions have been obfuscated and minimised right?
Have you tested that this operates as expected? It can't imagine it wouldn't, but the biggest worry is the i18n variables.

n3vu0r:
> Cool, so the big difference appears to be that variables internal to functions have been obfuscated and minimised right?

Exactly.

> Have you tested that this operates as expected? It can't imagine it wouldn't, but the biggest worry is the i18n variables.

Yes, I've build the image with these modifications. So far, I haven't seen any JS error in the browser's web console. The i18n variables should never be touched as they are global, otherwise it would be a bug of the minifier because we haven't enabled the `--toplevel` option.

The i18n JS files themselves are not processed by `do_js_compress()`. That would require modifications to the `gargoyle_header_footer.c` which expects line breaks after each i18n variable. I haven't found how haserl parses the JS files, might be similar. I wanted to estimate what we can additionally save if we would process these files as well. It would be 69 kB:

```sh
find package/ -path '*.js' -a -not -path '*/www/js/*.js' -print0 | du --files0-from=- -cb
find compressed_javascript/ -path '*.js' -a -not -path '*/www/js/*.js' -print0 | du --files0-from=- -cb
```

By doing so, I found JS files with invalid syntax, see #862.

lantis1008:
Apologies that it took me so long to get back to this. Thanks for your contribution :)

lantis1008 (approved):

PR #873, opened 2020-04-17
Gargoyle IPv6 Support

### LINKED ISSUE #834: Errors in IMQ and webmon
my router wr1200js
```
| Gargoyle version 1.11.X   | OpenWrt 18.06 branch               |
| Gargoyle revision ebad78b0| OpenWrt commit 90f6af5             |
| Built September 21, 2019  | Target  ramips/mt7621   
```
it show error when open qos

```
Fri Sep 27 19:52:54 2019 daemon.err uhttpd[4300]: Cannot find device "imq0"
Fri Sep 27 19:52:57 2019 daemon.err uhttpd[4300]: Cannot find device "imq0"
Fri Sep 27 19:52:59 2019 daemon.err uhttpd[4300]: Cannot find device "imq0"
Fri Sep 27 19:53:02 2019 daemon.err uhttpd[4300]: Cannot find device "imq0"
Fri Sep 27 19:54:04 2019 daemon.err uhttpd[4300]: sh: write error: Broken pipe
Fri Sep 27 19:54:20 2019 daemon.err uhttpd[4300]: Cannot find device "imq0"
Fri Sep 27 19:55:18 2019 daemon.err rmmod: module is not loaded
Fri Sep 27 19:55:18 2019 kern.info kernel: [564794.248709] 	IMQv4: Hooking IMQ after mangle on INPUT, pf=2, hooknum=1, priority=-149
Fri Sep 27 19:55:18 2019 kern.info kernel: [564794.256704] 	IMQv6: Hooking IMQ after mangle on INPUT, pf=10, hooknum=1, priority=-149
Fri Sep 27 19:55:18 2019 kern.info kernel: [564794.264739] 	IMQv4: Hooking IMQ after mangle on FORWARD, pf=2, hooknum=2, priority=-149
Fri Sep 27 19:55:18 2019 kern.info kernel: [564794.272841] 	IMQv6: Hooking IMQ after mangle on FORWARD, pf=10, hooknum=2, priority=-149
Fri Sep 27 19:55:18 2019 kern.info kernel: [564794.281075] IMQ driver loaded successfully. (numdevs = 1, numqueues = 1, imq_dev_accurate_stats = 1)
Fri Sep 27 19:55:18 2019 daemon.err insmod: module is already loaded - cls_fw
Fri Sep 27 19:55:18 2019 daemon.err insmod: module is already loaded - cls_flow
Fri Sep 27 19:55:18 2019 daemon.err insmod: module is already loaded - sch_hfsc
Fri Sep 27 19:55:18 2019 daemon.err insmod: module is already loaded - sch_sfq
Fri Sep 27 19:58:11 2019 kern.info kernel: [564967.599930] IMQ driver unloaded successfully.
Fri Sep 27 19:58:11 2019 kern.info kernel: [564967.652741] 	IMQv4: Hooking IMQ after mangle on INPUT, pf=2, hooknum=1, priority=-149
Fri Sep 27 19:58:11 2019 kern.info kernel: [564967.660731] 	IMQv6: Hooking IMQ after mangle on INPUT, pf=10, hooknum=1, priority=-149
Fri Sep 27 19:58:11 2019 kern.info kernel: [564967.668897] 	IMQv4: Hooking IMQ after mangle on FORWARD, pf=2, hooknum=2, priority=-149
Fri Sep 27 19:58:11 2019 kern.info kernel: [564967.677065] 	IMQv6: Hooking IMQ after mangle on FORWARD, pf=10, hooknum=2, priority=-149
Fri Sep 27 19:58:11 2019 kern.info kernel: [564967.685548] IMQ driver loaded successfully. (numdevs = 1, numqueues = 1, imq_dev_accurate_stats = 1)
Fri Sep 27 19:58:11 2019 daemon.err insmod: module is already loaded - cls_fw
Fri Sep 27 19:58:11 2019 daemon.err insmod: module is already loaded - cls_flow
Fri Sep 27 19:58:11 2019 daemon.err insmod: module is already loaded - sch_hfsc
Fri Sep 27 19:58:11 2019 daemon.err insmod: module is already loaded - sch_sfq
Fri Sep 27 19:58:31 2019 daemon.err uhttpd[4300]: sh: can't open '/tmp/tmp.sh': No such file or directory
Fri Sep 27 20:06:15 2019 daemon.err uhttpd[4300]: sh: you need to specify whom to kill
Fri Sep 27 20:06:17 2019 daemon.err uhttpd[4300]: sh: you need to specify whom to kill
Fri Sep 27 20:08:31 2019 daemon.info ddns_gargoyle[4907]: Checking whether update needed:
```

lantis1008:
**_Work In Progress_**
_I've opened this PR so that it can start receiving reviews and feedback early._
_More work will be pushed to this periodically, and I will let you know when _I_ think it is ready for merge._
Particularly @ericpaulbishop i'd appreciate a thorough review of the netfilter-match-modules. I'm pretty good at writing inefficient code, dereferencing NULL pointers, and other generally awful memory _faux pas_.

This PR adds IPv6 support to Gargoyle. I've been running this as my main internet gateway for about 3 months now in various stages of completion. Besides when I've been working on it, it has been very stable.

**Notable Changes:**
- netfilter-match-modules
  - All have been converted to xtables extensions instead of iptables. `timerange` has been enabled for all xtables. All other modules enabled for iptables and ip6tables only.
  - I have removed most of the legacy iptables 1.4 shim code (i think the error exit function is still there though, but could be removed)
  - Liberal use of `union ipany` which can store either an ipv4 or ipv6 to save space. Generally this means that an accompanying `int family` travels with the data to keep track of what we are dealing with.
  - bandwidth: Converted to entirely use of `string_map` instead of `long_map` to allow storage of IPv6 addresses without major modification to `tree_map.h`. 128 bit math is not easily supported so i did not go there. If this needs to be rethought for speed/efficiency, please let me know. Note that i did test sdbm collision with both ipv4 and ipv6 subnets occupying the same space and found no collisions in a basic home network setup, and did not encounter any in a larger test (not conclusive, ran out of RAM during computation lol)
  - bandwidth: Previously if you created two rules which reference the same ID, the rule would be rejected, but the iptables entry would still be created. Due to `ref_count` not being incremented, this would lead to a kernel panic once this rule was cleaned up. This behaviour has been changed so that now largely similar rules are rejected, unless one rule is ipv4 and one is ipv6. If the rule is rejected, the iptables entry is also never created. This means that a single ID can have both IPv4 and IPv6 addresses stored in its `string_map`. This saves complexity and memory in creating additional IDs, and also means that quotas can service both IPv4 and IPv6 devices at the same time (combined quotas anyone?)
  - libiptbwctl: bw_get and bw_set now take an additional parameter of `family` which is `2` for IPv4 and `10` for IPv6 (`AF_INET/NFPROTO_IPV4` and `AF_INET6/NFPROTO_IPV6` respectively). This means that bw data from earlier versions of Gargoyle are incompatible, i have not written a conversion module to handle this missing data. This would be relatively trivial (assume any data with no family is IPv4), but i don't think it is needed. Please let me know if you think otherwise.
  - webmon: Also outputs a new parameter `family` with same values as above.
  - conntrack: Also outputs a new parameter `family` which is either `IPv4` or `IPv6` in the GUI for filtering purposes
  - imq loading has been switched to a try/fail/retry method to try and cover the intermittent failure it has. I have not been able to track why it sometimes does not load. The added safety net seems reasonably robust, and in the event of a repeated failure attempts to cleanup after itself. This should hopefully fix the lack of NAT that some users were getting on a failed system, and also qos being _up_, but not actually working. Fixes #834 
- DHCP
  - Default operating mode is IPv6 Stateful mode. This allows us to fully control what addresses are handed out, as opposed to SLAAC which is just a crapshoot. If you want reliable firewall operation (restrictions, quotas etc), this is how it has to be. This does mean that Android devices do not get IPv6 in a Gargoyle network (SLAAC only, nice one Google 👎).
  - DUID is required to set a Static IPv6, but optional otherwise.
  - We only set the final 8 characters, and this applies to all prefixes that the router can delegate (GUA and ULA)
  - Host/IP mappings are now stored in /etc/config/dhcp instead of the old /etc/ethers and /etc/hosts. 
- Port Forwarding
  - Port forwarding doesn't make sense in the IPv6 space obviously. What you would instead want to do is open a port for a device. I could not think of a better place for this, so they share the same space as Port Forwarding.
- QoS
  - QoS rules now have the concept of an IP Family associated with the rule. If `src` or `dst` style rules are being written, the IP Family must be specifically set (it can't be _both_). Matches for just ports or protocols can be Family non-specific.
- restore_quotas
  - Now has a _dry run_ mode which instead of running the xtables commands, just prints them to console. Invoked with -p or -P
- Restrictions
  - Similar to QoS, now have an IP Family associated with the rule. If family is "any", can't use IP based rules (can still use local MAC Address matching)

**TODO:**
- **General: investigate why WAN6 does not always come up on its own. Sometimes needs a kickstart (ifup wan6), sometimes doesn't. May be a missing call somewhere.**
- General: cleanup tab vs space indentation. Got my editors mixed up and i've caused quite a mess. Any suggestions for cleaning this up without ruining git history much appreciated.
- General: cleanup duplicated code (some things could be done better and more efficiently)
- General: cleanup use of _full format_ ipv6 addresses. We should prefer canonical where possible. This also means that we compare apples to apples and don't end up considering `2001:db8::` differently to `2001:db8:0000::` etc.
- Connection -> Basic: PPPoE untested. Need guinea pigs who are advanced users.
- **Firewall: More generally support the _negative netmask_ notation for IPv6 to handle dynamic global prefixes from ISPs.**
  - QoS should already support this
  - bandwidth/quotas need slight modifications to their subnet comparison modules to support this. No plan of attack yet.
  - Webmon/Weburl probably need similar modifications
  - Restrictions should be fine out of the box, except for the bits which rely on quotas.
  - Need to accept the negative subnet notation throughout the GUI (where appropriate)
- gargoyle_header_footer: `current_lan_ip6` and `current_wan_ip6` need to be fixed so they can fetch from UCI as well as just the interface. Tricky part is that these interfaces can have > 1 IP listed in UCI, so we need to bring them in as a list rather than static options. No big deal, just haven't done it yet.
- ewget: support retrieving from ipv6. patchset mostly done.
- qosmon: ping target IPv6 capability not done. 
- plugin-gargoyle-openvpn: Not done (low priority, not sure of the benefits here)
- plugin-gargoyle-ddns: Not yet done (relies on ewget)
- plugin-gargoyle-tor: Not done (low priority, not sure of the benefits here)
- plugin-gargoyle-usb_storage: Have not checked any of the file sharing over IPv6. In theory should be automatic.
- plugin-gargoyle-webcam: Not done (don't have any webcam so can't test)
- plugin-gargoyle-usb_printer: Not done (don't have any printer so can't test)
- plugin-gargoyle-minidlna: Not yet done
- plugin-gargoyle-ping-watchdog: Not yet done
- plugin-gargoyle-pptp: Not done (low priority)
- plugin-gargoyle-adblock: Not done (low priority)

Pictures:
Status -> Overview
image
Status -> Bandwidth Usage
image
Connection -> Basic
image
Connection -> DHCP
image
Firewall -> QoS
image
Firewall -> Quotas
image
Firewall -> Restrictions
image


lantis1008:
Commentary from @obsy , @pbix and @ispyisail certainly most welcome and encouraged. As well anyone else who would like to chime in.

obsy:
Wow, very huge changes :)

ispyisail:
I guess we wait a bit for Eric to comment?

> plugin-gargoyle-openvpn: Not done (low priority, not sure of the benefits here)

What does this mean? OpenVPN IPv4 will still work?

lantis1008:
I don't know if he has seen it yet and is digesting it, or hasn't come across it. 
I'll send him an email in a few days time if nothing before then. 

Yes it will still work over IPv4, I just won't (at least initially) put any effort into pushing IPv6 over the tunnel. Partly because I don't think it is overly necessary, and mostly because I don't fully understand how it works yet.

pbix:
How about QOSMON?  It is currently uses a IPv4 ping target.

lantis1008:
QOSMON will get an upgrade, but so far I have left the ping target logic as is. 
I'll add it to the list so I don't forget, thanks!

lantis1008:
Off the top of my head, that one should be straight forward.

livepu:
Very powerful

lantis1008:
I've heard from Eric and he is happy to defer to my judgement call on this one.
Does anyone have any strong objections to me merging this into Master?
Does anyone have any "show stoppers" that should be resolved first?

This will allow us to get test versions in more peoples hands and hopefully uncover bugs that i have not thought of.

@obsy @ispyisail @pbix

obsy:
Do it.

livepu:
I support testing, merge quickly

PR #877, opened 2020-04-29
Updated URL list

aruljohn:
I updated the URL list with a more IP-friendly URL for IP lookup regarding the aruljohn.com entry. The current URL for aruljohn.com requires the webpage to be parsed to get the IP address. Visiting the new URL aruljohn.com/ip/ returns just the IP address.


lantis1008:
Thank you very much for addressing this @aruljohn .
It was only recently that I discovered that the IP has some default value and is runtime evaluated. It explains the error.

obsy:
Maybe url list should moved to config file?

lantis1008:
@obsy Yes, i will investigate this as part of a separate PR i think.

PR #882, opened 2020-05-25
Ignore false detection of bootloader information in tplink ramips/mt7…

obsy:
…6x8 images

Images for ramips/mt76x8 contain 0xff, not 0x00


lantis1008:
Thanks, sorry for the delay

PR #901, opened 2020-12-08
Add cron task translation file for ZH-CN, correct few translate

crackself:
There's one need translate, but I didn't find the right file to modify. If someone know how to, please give some tips.
 The `World（Defualt）` should be `全球 (默认)`, and `undefined` should be `未指定`


lantis1008 (approved):

lantis1008:
I have not made the country list translatable to save space.
https://github.com/ericpaulbishop/gargoyle/blob/master/package/gargoyle/files/www/data/countrylist.txt

I'm open to trying to translate only the _World (Default)_ entry if you wanted to submit some kind of patch? I don't think it is appropriate to translate all country names.

crackself:
> 
> 
> I have not made the country list translatable to save space.
> https://github.com/ericpaulbishop/gargoyle/blob/master/package/gargoyle/files/www/data/countrylist.txt
> 
> I'm open to trying to translate only the _World (Default)_ entry if you wanted to submit some kind of patch? I don't think it is appropriate to translate all country names.

Surely，translate all country names is unnecessary. I will try to patch the default entry.

lantis1008:
Possibly somewhere here:
https://github.com/ericpaulbishop/gargoyle/blob/master/package/gargoyle/files/www/js/basic.js#L4131-L4149

You could add a special condition to check if the translation is available and substitute it. Otherwise provide the original text.
Thanks for looking

PR #902, opened 2020-12-08
plugin adblock: removing error display

obsy:
Fixes for this if plugin tor is not installed:
```
root@Gargoyle:/usr/lib/adblock# sh ./runadblock.sh 
uci: Entry not found

```


lantis1008 (approved):
Thank you

PR #903, opened 2020-12-23
QR Code Support for Gargoyle

n3vu0r:
This adds a new package `plugin-gargoyle-qr-code` to generate QR codes for Wi-Fi configurations, router access, and webcam snapshpts/streams as downloadable SVGs or printable documents. It provides two panels, a QR code editor and a viewer. The editor lets you individually configure home/guest Wi-Fi, web access, and webcam snapshot/stream applications. It provides checkboxes for what should be included in the viewer, a text area for custom comments supporting Markdown formatting, and checkboxes of whether the viewer should embed this application in the local or remote login page.

qr_code-1
## Wireless Network > Home Wireless Network
qr_code-2
## Router Access > Web Access > Local Web Access
qr_code-3
## Local Login Page > Local Web Access
qr_code-4

---

## Commits

 1. Fine tune light and green themes. Could be a separate PR like the next commit.
 2. The basic connection page triggered an error when loading the new UCI values directly after successfully saving them. It was accidentally saved as two-dimensional row-column array of a single column. Saving to configuration file still worked but the new value stored in `uci`/`uciOriginal` caused `resetData` to fail.
 3. The JS to determine whether a page is viewed locally or remotely was duplicated in `access.js` and `webcam.js`.  Now it is shared as template, supports IPv6 addresses, and is also used in `qr_code.js` and its login hook `090-qr_code.js`.
 4. Since the viewer is included in the plugin page and in the login page it is a template with shared JS and CSS. For this I extended the login hook script to be able to include shared resources configurable via UCI, see `saveChanges` in `qr_code.js`.
  5. The editor and viewer use option groups. I put this functionality in `common.js`.
  6. Functionality for text areas, also put in `common.js`.
  7. I thought it is a good idea to somehow abbreviate `document.getElementById` for code size and right drift.
  8. The actual plugin with its two MIT licensed dependencies:
        * [QR Code Generator](https://github.com/nayuki/QR-Code-generator): Correctness and available in many languages.
        * [Markdown Parser](https://github.com/adamvleggett/drawdown): One of many trying to be very small. It seems to be reliable while typing, after a small modification.


lantis1008:
Thanks very much for sending this. It seems quite well thought out with lots of great functionality.
I'd like to take the time to review it properly, so please give me a bit of time.
I'm pretty happy with the first 3 commits, after that I need to take a proper look.

After merge, i'd love to see this extended to cover the Wireguard plugin use case as well if possible.

package/plugin-gargoyle-i18n-Polish-PL/files/www/i18n/Polish-PL/qr_code.js
```diff
@@ -0,0 +1,47 @@
...
+QrCode.DownloadQrCode="Pobierz kod QR";
+QrCode.QrCodeEditor="Edytor kodów QR";
+QrCode.Application="Aplikacje";
+QrCode.SelectApplication="Wybierz aplikację";
+QrCode.Wifi="Sieć bezprzewodowa";
+QrCode.RouterAccess="Dostęp do routera";
+QrCode.WebAccess="Dostęp do sieci";
+QrCode.Webcam="Kamera internetowa";
+QrCode.WebcamSnapshot="Migawka z kamery internetowej";
+QrCode.WebcamStream="Istrumień z kamery internetowej";
```
obsy, line 36: Typo: "Strumień z kamery internetowej", not "Istrumeń..."

n3vu0r: Thanks!

n3vu0r:
Glad you like it. Take all the time it takes. I'll have a look at the wireguard plugin, this seems interesting.

When rebasing I forgot the two comments in 6681cba, they are added with d11e141 instead.

package/gargoyle/files/www/js/access.js
```diff
@@ -591,40 +576,43 @@ function updateVisibility()
 		remoteHttpsElement.value = "";
 		remoteHttpElement.value = "";
 	}
-	document.getElementById("noip6redi").style.display = remoteWebProtocol != "disabled" ? "block" : "none";
+	document.getElementById("noip6redi").style.display = remoteWebProtocol == "disabled" || (remoteWebProtocol == "https" && localWebProtocol == "redirect") ? "none" : "";
```
lantis1008, line 579: Please use "block" and "none".
Setting to "" instead of "block" will not override any styles that may be applied by other CSS (i.e. not directly to the element) potentially leading to less defined behaviour. This is also more consistent with the remainder of Gargoyle

package/gargoyle/files/www/js/access.js
```diff
@@ -591,40 +576,43 @@ function updateVisibility()
...
-			'<ol>' +
-				'<li>' + accessStr.ReverseProxySelect + ' <i>HTTP & HTTPS</i> ' + accessStr.ReverseProxySaveCh + '.</li>' +
-				'<li>' + accessStr.ReverseProxySwitch + ' <i><span id="reverse_proxy_switch"></span></i>.</li>' +
-				'<li>' + accessStr.ReverseProxySelect + ' <i><span id="reverse_proxy_select"></span></i> ' + accessStr.ReverseProxySaveCh + '.</li>' +
-			'</ol>' +
-		'</span>' +
-		'<p>' + accessStr.AccessChain + ':</p>' +
-		'<p><span id="access_chain"></span></p>' +
-	'</span>';
+	`<span id="lockout" class="col-xs-12">
```
lantis1008, line 601: Please don't use backtick templating. This does not seem to play nicely with ./i18n-scripts/localize.py.
Alternatively, please fix localize.py to correctly address backtick template variables.

package/gargoyle/files/www/js/access.js
```diff
@@ -636,8 +624,9 @@ function updateVisibility()
 		document.getElementById("reverse_proxy_select").innerText = allOptionValueHash[accessWebProtocol];
 
 		// Web access trace of domains and/or IP addresses from client, possibly over reverse proxy, to server.
+		var ipHost = function(ip) { return ip.includes(":") ? `[${ip}]` : ip; };
```
lantis1008, line 627: Backticks,consider revising (but no localize impact)

package/gargoyle/files/www/js/common.js
```diff
@@ -12,6 +12,33 @@ var TSort_Classes = new Array ('odd', 'even'); // table sorting zebra row suppor
 var menuState = []; //nav Menu State variable
 var navTimer; //timeout for restoring menu
 
+// Abbreviation.
+function byId(id)
```
lantis1008, line 16: Would like this more if we added an optional second parameter to set the document variable to something else (as is done in some other parts of Gargoyle).
e.g. function setChildText in common.js

package/gargoyle/files/www/js/common.js
```diff
@@ -1283,6 +1328,68 @@ function setAllowableSelections(selectId, allowableValues, allowableNames, contr
...
+	const optgroup = document.createElement("optgroup");
+	optgroup.label = label;
+	optgroup.hidden = hidden;
+	optgroup.disabled = disabled;
+	select.appendChild(optgroup);
+	return optgroup;
+}
+
+// Add option and unhide its group if present.
+function addOption(selectOrOptgroup, text, value = "", selected = false, hidden = false, disabled = false)
```
lantis1008, line 1343: We already have addOptionToSelectElement. Can these be consolidated while maintaining compatibility with existing code? No sense having two. If not, this function should be made specifically to add to OptGroups only (and documented as such) or moved to qr_code_commn.js.

package/gargoyle/files/www/js/common.js
```diff
@@ -1283,6 +1328,68 @@ function setAllowableSelections(selectId, allowableValues, allowableNames, contr
...
+
+// Get selected option group.
+function getOptGroup(select)
+{
+	const option = getOption(select);
+	return option ? option.parentNode : null;
+}
+
+// Get selected option.
+function getOption(select)
```
lantis1008, line 1373: We already have getSelectedValue. Same question as above.

package/gargoyle/files/www/templates/client_server_template
```diff
@@ -0,0 +1,47 @@
+
```
lantis1008, line 1: Please remove leading blank line. If blank line separator needed, add on the parent page.
Please check all instances where you have called this template and add the blank line there to create proper separation (if/as needed).

package/plugin-gargoyle-qr-code/files/etc/config/qr_code_gargoyle
```diff
@@ -0,0 +1,44 @@
+
```
lantis1008, line 1: Please remove leading blank line

package/plugin-gargoyle-qr-code/files/www/hooks/login/090-qr_code.js
```diff
@@ -0,0 +1,97 @@
...
+ * version 2.0 with a special clarification/exception that permits adapting the program to
+ * configure proprietary "back end" software provided that all modifications to the web interface
+ * itself remain covered by the GPL.
+ * See http://gargoyle-router.com/faq.html#qfoss for more information
+ */
+
+// ECMAScript 6 (ES6) Strict Mode Code
+"use strict";
+
+addLoadFunction(() =>
```
lantis1008, line 12: "Arrow Functions" irk me a little. I won't say no, but please carefully consider their use.
This is probably an old fashioned comment. 😄

package/plugin-gargoyle-qr-code/files/www/hooks/login/090-qr_code.js
```diff
@@ -0,0 +1,97 @@
...
+{
+	// Whether shared login hook resources of plugin are loaded.
+	const pkg = uciOriginal.get("gargoyle", "qr_code", "pkg");
```
lantis1008, line 15: Use of `const` where `var` is likely sufficient in many locations. Save those bytes 2 at a time!
Again I won't say no, but consider their use.

package/plugin-gargoyle-qr-code/files/www/templates/qr_code_viewer_template
```diff
@@ -0,0 +1,55 @@
+
```
lantis1008, line 1: Please remove leading blank line

lantis1008 (changes requested):
Please see inline comments. Additionally:
6681cba - Novel. I think in general keeping login hooks separated makes sense, but also no sense duplicating code. This could be useful in the future.
2313312 - Seems to be a little bit of duplicated code with some existing functions for not much gain. I wonder if OptGroups could be replaced by just removing them altogether, or maybe with a second dropdown box (category and type). What you've done is neat, but the duplication may not be worth the minimal gains? There are comments inline, if you can't address the duplication easily/effectively, an option might be to drop this functionality? Your call I won't block it just for this reason.

Functionality wise, very easy to use and works great. Good job! As i said, would love to see Wireguard options added (either in v1 or as an additional feature later). None of my comments were deal-breakers, i'd pretty happily merge it as is, however if you could take a quick review that would be great.

Thanks and Merry Christmas

package/gargoyle/files/www/js/access.js
n3vu0r: Will do. At some places I use `inline-block` in CSS. Do I use `"inline-block"` then in JS or stick to `""`?

package/gargoyle/files/www/js/common.js
n3vu0r: Done.

package/gargoyle/files/www/js/access.js
lantis1008: Yes use the appropriate display setting as needed

package/plugin-gargoyle-qr-code/files/www/hooks/login/090-qr_code.js
n3vu0r: I see that, especially here with a multi-line function in almost-global scope. Will change it in these places. But I would suggest to use arrow functions at least in places with a lot of one-liners like adding a set of event listeners or with array combinators:

`map(id => byId(id))` compared to `map(function(id) { return byId(id); })`.

package/plugin-gargoyle-qr-code/files/www/hooks/login/090-qr_code.js
n3vu0r: It seems `const`s survive terser and there is no switch to get rid of them. I would prefer to go with `let` then because of block scope and no hoisting.

package/gargoyle/files/www/js/common.js
n3vu0r: I think it's currently a bit too different from existing functions (DOM object vs ID string as arguments) and unless used elsewhere it's better too keep `common.js` small.

package/gargoyle/files/www/js/access.js
n3vu0r: Removed all backtick templates.

---
I haven't fully understood `localize.py` but it seems to work fine with backtick templates. It replaces `${jsObj.Variable}` with `${"Localized Text"}` which is totally valid and respects original escaping. The syntax is `${ expression }` where expression is not limited to variable names.

n3vu0r: But it can be problematic as soon as there is a backtick within the localized text.

n3vu0r:
I hope the clean up went in the right direction, let me know otherwise. I'm currently undecided regarding option groups. It's possible without them but the viewer selection can get messy with up to three kinds of URLs (local/remote/indirect) which themselves contain all http/https and IPv4/IPv6 combinations. Having two select elements is mostly overkill in cases where only one category/group is offered and requires more user clicks.

It seems quite easy to add WireGuard support. The QR code is simply the text content of a configuration file. And I think it makes only sense to offer QR codes for the list of allowed clients where the public and private keys are managed by Gargoyle. Will give it a try tomorrow.

lantis1008:
Looks great thanks. I'll merge this now and you can push something for Wireguard when you're ready.

I agree that single line functions are probably ok to use arrow functions.

lantis1008 (approved):

PR #907, opened 2021-01-01
Fix client/server template with DHCP on WAN.

n3vu0r:
When using DHCP on WAN side, `uci -q get network.wan.ipaddr` is empty and `uci -qp /tmp/state get network.wan.ipaddr` has to be used instead. But this can have timing problems so I inverted the test logic and went with `uci -q get network.lan.ipaddr` which I think is permanently set.

Happy new year!


lantis1008 (approved):
Thanks. 
I'm hoping to remove all references to /var/state in the future as there is a more defined method of retrieving such information now.

PR #910, opened 2021-01-06
build: Fix npm trying to install to privileged location

lantis1008:
Which may not be possible for all users on all systems.
Based on feedback from forums:
https://www.gargoyle-router.com/phpbb/viewtopic.php?p=58029#p58029

Waiting on further feedback of confirmed fix before merging.


lantis1008:
Confirmed fixed: https://www.gargoyle-router.com/phpbb/viewtopic.php?p=58079#p58079

PR #913, opened 2021-01-28
plugin e-mail: some improvements

obsy:
- set the execution bit (group, other) to /usr/lib/gargoyle/email.sh
- reverse ssl library detection logic

msmtp can be installed from the openwrt repository and compiled with libgnutls. In this case, detection of the ssl library will fail (msmtp --version: "TLS / SSL Library: GnuTLS"). Reverse logic and detect msmtp without SSL support.

- by explicitly calling msmtp instead of sendmail

openwrt mstmp does not provide a link to sendmail

See old topic: https://www.gargoyle-router.com/phpbb/viewtopic.php?p=46337#p46337


lantis1008:
Hey, apologies for the delay, I'll merge this when I return from holidays :)

lantis1008 (approved):

PR #914, opened 2021-02-03
Remove var state usage

lantis1008:
/var/state is deprecated upstream, and has largely been replaced with `ifstatus` and several wrapper functions in `/lib/functions/network.sh`.
Move Gargoyle to using ifstatus for header/footer generation, and wrapper functions elsewhere as appropriate.

This has not been thoroughly tested in pppoe use cases, or usb internet usage. Would appreciate any feedback on these in particular.

gargoyle_header_footer now relies on libjson-c, which we were already pulling in anyway as a dependency for other items. Upstream master appears to be moving more towards jansson, and we may follow this in time.
These changes could be written without a json library, but it would be a bit more challenging than i was prepared to undertake at the time.


lantis1008:
@obsy do you have any USB modems that are in "stick" mode? I want to understand what needs to change with this file:
https://github.com/ericpaulbishop/gargoyle/blob/master/package/gargoyle/files/etc/hotplug.d/iface/99-qmincm

The whole usb internet part is something i have not tested at all.

obsy:
I ... don't remember what it was for. Nothing uses modem interface specifically, so this probably can be removed.

lantis1008:
I'm going to merge this now. There is some outstanding items in DDNS. I'm going to leave this for now as @ericpaulbishop was going to do some work in this area and I don't want to create any conflicts.

PR #933, opened 2021-08-28
ath79: Add support for TP-Link Archer C60 v3

MariosK239:
I have tried to add support for the tplink archer c60 v3 router. It is very similar to the v1 and v2 versions. I am not sure if i have done anything wrong, but i built an image and it seemed to work fine.


lantis1008 (approved):
Thanks for your contribution

PR #936, opened 2021-12-27
Update Edgerouter X profile image regexp to match OpenWrt 19.07 naming

aimacintyre:
See image names for 19.07.8 at
https://downloads.openwrt.org/releases/19.07.8/targets/ramips/mt7621/


lantis1008:
Thanks for your contribution!

PR #938, opened 2022-01-02
ramips.mt7621: add support for Netgear R6220

aimacintyre:
In OpenWrt the Netgear R6220 has been supported by the mt7621
profile of the ramips target since 18.06.

A factory image built from Gargoyle commit cf669bf0 with these
changes has been successfully installed on an R6220 via the OEM web
interface and run; a sysupgrade has also been successfully completed.


lantis1008:
Thanks for your contribution!

PR #943, opened 2022-01-30
build: Properly set num_build_thread_str in auto mode

jan-kiszka:
As this var was not set, we only built in single-core mode when "auto"
was selected.

Signed-off-by: Jan Kiszka <jan.kiszka@web.de>


jan-kiszka:
BTW, the amount of code duplication between `build.sh` and `rebuild.sh` looks... improvable.

lantis1008 (approved):

PR #957, opened 2022-09-20
Add newline at end of config.in files in OpenVPN package

aimacintyre:
Without the newline at the end of each of Config-mbedtls.in and Config-openssl.in, the make menuconfig command in build.sh for "make custom" fails with errors about the missing newlines and complaining about syntax; the build process then continues without stopping until all the default ath79 images have been built.

The respective .in files in the current master branch (for 1.13.x) have the newline at the end of them, so make menuconfig works as expected there.


lantis1008:
Thanks again!

PR #958, opened 2022-09-23
ipq806x: add extra wifi driver and firmware packages, add support for Linksys EA7500 v1

aimacintyre:
Commit cca11a2:
While the ct (Candela Tech) driver and firmware packages are the
OpenWrt default, and should remain the Gargoyle default, some users
report better results with the original (aka non-ct) driver and
firmware packages.  As these are kmods, the OpenWrt packages can't
be used for Gargoyle so build them to give Gargoyle users the
option of installing them in place of the default ct packages.

Additionally there are now several variants of the ct firmware
packages and smallbuffers variants (which reduce memory
consumption) of the both ct and non-ct driver packages so build
them as well.

Commit 27eb818:
OpenWrt 21.02 added support for the Linksys EA7500 v1, which is
similar to the Linksys EA8500 except for having 256MB of RAM
(instead of 512MB) and 3 antennas (instead 4 antennas).

### DIFF
--- a/targets/ath79/profiles/ath10k-large/config
+++ b/targets/ath79/profiles/ath10k-large/config
@@ -3124,16 +3124,16 @@ CONFIG_PACKAGE_ath10k-board-qca988x=m
 # CONFIG_PACKAGE_ath10k-firmware-qca4019-ct-htt is not set
 # CONFIG_PACKAGE_ath10k-firmware-qca6174 is not set
 # CONFIG_PACKAGE_ath10k-firmware-qca9377 is not set
-# CONFIG_PACKAGE_ath10k-firmware-qca9887 is not set
+CONFIG_PACKAGE_ath10k-firmware-qca9887=m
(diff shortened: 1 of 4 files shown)


lantis1008:
Thanks for this. 
I agree with the proposal to build (but not include by default) the other variants of ath10k drivers+firmware. 

Do you have time/effort available to make a similar patch for the ath79-ath10k/ath10k_large profiles as well? Otherwise I will tackle this at my next available opportunity.

aimacintyre:
I've added commit d1e9f49 with the config changes for the two ath79 ath10k profiles; both compile tested however I note that with the ath10k profile only the C7 v1 images are completed.  I think the others fail the image size checks.  I've attached the make log for that profile in case it's helpful.  All the ath10_large images are completed.
[make.ath79_ath10k.log.gz](https://github.com/ericpaulbishop/gargoyle/files/9634507/make.ath79_ath10k.log.gz)

lantis1008:
Thanks! Will review non-building profiles soon

PR #960, opened 2022-12-26
fixes/extension to Config-gargoyle.in to get all required packages at least buildable with minimal diffconfig

aimacintyre:
2 patches to Config-gargoyle.in:
- 5db53f0: add machinery so that the profile meta-packages default to "M" in the config after a `make defconfig`
- a4ec89d: rejig and extend the machinery to get all other necessary packages to "M" in the config after a `make defconfig`

da5b6a8 should be ignored, as that is just bringing the base_on_openwrt_2102_ath79buildfix branch up to match the current state of base_on_openwrt_2102; in hindsight I should have deleted the branch and started a new one but I'd hoped the additional commits might appear in the history of pull request 959 (a vain hope I should have realised as that pull request is closed).


aimacintyre:
Added bc71613 which is a minor cleanup to the curl package Config.in to avoid the unwanted appearance of LIBCURL_WOLFSSL when `make defconfig` is run - explicit overrides need to be in a diffconfig file otherwise to remove the LIBCURL_WOLFSSL reference and enable the wanted LIBCURL_OPENSSL instead.

lantis1008:
Just pulled this into my local.
If i manually set the 3 meta packages to _is not set_, then run defconfig, they don't get set to _m_.
Is this the expected behaviour?

Merry Christmas by the way :)

aimacintyre:
If you make the meta-packages _is not set_ in a diffconfig file then yes I'd expect that to override the Config-gargoyle.in "m" settings as the diffconfig acts as an override.  It's not always straightforward though, because as far as I can tell packages with DEFAULT references (e.g. opkg) seem to only get downgraded from "y" to "m" in the final .config with the package _is not set_ in a diffconfig file which is a bit annoying even if they don't make it into the device images.

Aside: it might be useful to introduce a list of unwanted/useless packages not to copy into the build directory - okpg & wpad-basic-wolfssl are 2 that I know of that need the diffconfig overrides to get them set to "m" and so are in the collection of packages for each target.  Alternatively it might be worth a post-defconfig patching run to make such packages explicitly _is not set_ in the generated .config.  The DEFAULT package references don't seem to affect the outcome of the actual build if the respective packages are _is not set_ in the .config.

I suspect that there might be a couple of other BUSYBOX config defs that need to be added to what's in the second patch - I just copied what you'd included in your original Config-gargoyle.in proposal in a comment to pull request [959](https://github.com/ericpaulbishop/gargoyle/pull/959) but I think I've seen some others that might be candidates in a couple of the target profile configs.

There are also a number of kernel and build parameters that need selection on a per target basis which I think are best handled in per-profile diffconfig files, particularly items that need disabling to reduce kernel size for some targets (the older targets like ath79 and ramips especially).  As well as the devices and the respective profile meta-packages...

A Merry Christmas to you too and I hope the New Year is good for you!

PR #962, opened 2023-01-01
`make menuconfig` related config fixes

aimacintyre:
The first 4 commits (bf6b32b, c973d6d, f38e7b3, 4bc7f7e) all relate to an issue where `make defconfig` correctly sets all the expected package selections but then `make menuconfig` won't display the profile meta-package selections or preserve them in the output .config.  These changes represent the least invasive changes I was able to identify that rectified the `make menuconfig` behaviour and with them in place I was able to generate an ath79 config from a diffconfig with the profile meta-packages specified in device packages lists (as discussed in recent comments to #959) and then select an additional device and specify one of the profile meta-packages as a device package and have `make menuconfig` retain everything in the base .config and include the extra config items for the additional device in the final .config.

Commit e666b5b is a very minor, non-critical, cleanup to the gargoyle-firewall-util package to scrounge back a smidgin more flash storage on devices without plugin-gargoyle-pptp installed.

### DIFF
--- a/files-generic/config/Config-gargoyle.in
+++ b/files-generic/config/Config-gargoyle.in
@@ -93,6 +93,8 @@ menuconfig USE_GARGOYLE_PROFILE_PKGS
 			select PACKAGE_plugin-gargoyle-webshell
 			select PACKAGE_plugin-gargoyle-wifi-schedule
 			select PACKAGE_resolveip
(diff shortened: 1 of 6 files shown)


lantis1008:
Brilliant, thanks for solving the recursive issue. These are all good changes.

PR #965, opened 2023-01-15
Fixed displaying DNS addresses

obsy:
Removing exactly the IP address, not a string of characters starting with that address. If the DNS address is in the same address class as LAN, it is removed from the list but should stay.

For example
```
echo "192.168.1.100" | grep -v '192.168.1.1'
(return nothing)

echo "192.168.1.100" | grep -v '192.168.1.1$'
192.168.1.100
```
Also please add these changes to other branch.

### DIFF
--- a/package/gargoyle/files/www/overview.sh
+++ b/package/gargoyle/files/www/overview.sh
@@ -60,7 +60,7 @@
 
 	lan_ip=$(/sbin/uci get network.lan.ipaddr 2>/dev/null)
 	if [ -n "$lan_ip" ] ; then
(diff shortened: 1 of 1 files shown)


lantis1008:
Hey, we've got this picked up in https://github.com/ericpaulbishop/gargoyle/commit/09aef320909d614ec73ab41763db72eb0c20cfe2
I can merge to master if you wish, but it will cause a conflict later when i go to bring in 22.03 branch. Not a big deal it can be solved easily.

obsy:
So no, keep this in 22.03 branch.

PR #967, opened 2023-01-21
new targets: ramips.mt76x8 and ipq40xx

aimacintyre:
New ramips subtarget mt76x8: a variety of lower end devices, some still commercially available at time of writing.

New ipq40xx target: a mid-range SoC family with many devices reasonably well resourced and currently commercially available.

I've test run devices in both targets (and run a build on a GL-B1300 as a daily driver) and public experimental builds of both have had at least several 10s of downloads (close to 100 downloads across 2 builds for mt76x8, about 60 across 2 builds for ipq40xx) although with limited feedback :-(.


targets/ipq40xx/profiles/default/profile_images
```diff
@@ -0,0 +1,19 @@
+_jalapeno-squashfs-
```
lantis1008, line 1: Will accept as is, but in a subsequent PR can you please add the manufacturer string to the front of this? For consistency more than anything else. Thanks!

targets/ramips/profiles/mt76x8/profile_images
```diff
@@ -0,0 +1,19 @@
+gl-mt300n-v2-squashfs-
```
lantis1008, line 1: Same comment as ipq40xx

lantis1008 (approved):
Beautiful

targets/ipq40xx/profiles/default/profile_images
lantis1008: Just when you get time or are sending another round of changes. Not urgent!

PR #974, opened 2023-02-09
add bcm27xx target, USB ethernet "extras" package

aimacintyre:
First: minor non-critical clean-up for the plugin-gargoyle-usb-storage-noshare package: kmod-usb2 is not required as a dependency and better left to device image recipes to include only on those devices where it's required.

Add a Gargoyle "extras" package as a container for meta-packages collecting various packages of related utility together for convenience of inclusion in target configs as well as for later installation on live systems.  The only meta-package included in this commit is for USB ether net drivers - which are sometimes useful on regular routers but most useful on single board computer type devices with only a single ethernet port but spare USB ports.

Add the bcm27xx target which includes all RaspberryPi variants currently supported by OpenWrt (Pi 0, Pi B, Pi 2, Pi 3 and Pi 4).


targets/bcm27xx/patches/50-add-eth1-to-board-config.patch
```diff
@@ -0,0 +1,10 @@
+--- a/target/linux/bcm27xx/base-files/etc/board.d/02_network	2021-08-21 22:15:47.655631601 +1000
```
lantis1008, line 1: Could this be handled instead in the same way as x86 ethernet?
https://github.com/ericpaulbishop/gargoyle/blob/base_on_openwrt_2203/targets/x86/patches/50-x86setup.patch

Only a suggestion, your call. It's a neat patch either way.

targets/bcm27xx/profiles/bcm2711/diffconfig
```diff
@@ -0,0 +1,4 @@
+CONFIG_TARGET_bcm27xx=y
```
lantis1008, line 1: A few diffconfigs included, is this intentional? I think they should be removed for this PR?

aimacintyre: Including the diffconfigs was unintentional - they should be removed (will do this when I'm back from travels late this week).

targets/bcm27xx/patches/50-add-eth1-to-board-config.patch
aimacintyre: I'm sure the network change could be handled the same way as x86 and I looked at doing that when I assembled the bcm27xx target originally against 1.13.x/OWrt19.07.  I can't remember where I found the basis of the technique in this patch (most likely another OpenWrt target 02_network file I think) but I used it because it was so much simpler and worked for most bcm27xx boards (which have an onboard ethernet port, even if it's USB attached, which is reliably enumerated as eth0 before any extra USB interfaces are added) without appearing to interfere with use of LTE dongles.  It worked on 1.15.x too in my limited testing so I kept it.  Based on what I've read, not yet having been able to test, the PCI NIC on the DFRobot CM4 router board is enumerated as eth1 too so I expect that setup to work the same as the Pi4 with a USB ethernet dongle.  I don't have any of the WiFi only devices (e.g. Pi0W) to test to see what would be needed for them.

The x86 ethernet situation undoubtedly requires more sophistication/complexity; so far I don't think that's needed for bcm27xx so would prefer to keep this patch until it proves insufficient.

lantis1008:
No worries. If you remove the diffconfigs I can merge. No rush!

aimacintyre:
Bah!! Git never does what I expect :-( - I'll close this and redo the PR.

PR #975, opened 2023-02-14
Port ntfs3 to 22.03-based Gargoyle

lukasz1992:
Hello,

Even if ntfs3 is present only in snapshot, I decided to push update anyway - as I will use it.

Source code comes from latest 5.15 kernel (you can download it from https://kernel.org/)
0004-0008 patches are already prepared, 0009 was prepared by myself, and 0001-0003 is fix for BE machines

I tested build and confirmed that kernel module is being loaded on x86 QEMU.

### DIFF
--- a/package/ntfs3/Makefile
+++ b/package/ntfs3/Makefile
@@ -1,0 +1,36 @@
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/kernel.mk
+
+PKG_NAME:=ntfs3
+PKG_VERSION:=5.15.93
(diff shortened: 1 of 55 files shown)


lukasz1992:
NTFS compression is rarely used, but we could add support for reading compressed files with about 7% increase of kernel module size - guessing that this is 20kB difference for uncompressed binary.

If we want it, need to add additional line 33 in Makefile:

		CONFIG_NTFS3_LZX_XPRESS=y \

lantis1008:
Thanks, i'll review this and get back to you

lantis1008:
Works great, thanks for your contribution

PR #980, opened 2023-02-26
Update wifi settings

lukasz1992:
I propose quite big set of changes:

The first one is handling Wifi6 on 2.4GHz.

I think that some options should be dropped:
- WiFi 2, because OpenWrt 21.02 and newer disable 802.11b by default
- ad-hoc - I think nobody even test it
- WPA1 - I could not think about any example of WPA1 AES device, which could not handle WPA2 AES;
- WEP - it is very weak and breaks WiFi4 (hostapd will disallow HT)
- WiFi 3 (802.11g and 802.11a) - I think that it is a good time to stop supporting ancient devices.

Some small fixes also attached:
- added channel 149 for 160MHz (in theory it is okay, but it would not probably appear often due to regulations)
- fixed wifi based on broadcom drivers (broken by introducing WiFi6)
- fixed invalid output from iw (when mac80211 supports 802.11ah)

I prepared branch for those all and I am open to discuss.
I also gave a chance to stay at WiFi3, when (and only when) radio does not support WiFi4.

If any special case needs handling, remember that Gargoyle here works as overlay for OpenWrt, you can still tweak these settings by editing /etc/config/wireless by your own.

### DIFF
--- a/package/gargoyle/files/usr/lib/gargoyle/cache_basic_vars.sh
+++ b/package/gargoyle/files/usr/lib/gargoyle/cache_basic_vars.sh
@@ -95,7 +95,7 @@ print_mac80211_channels_for_wifi_dev()
 	# however, as far as I can tell there is no other way to get max txpower for each channel
 	# so... here it goes.
 	# If stuff gets FUBAR, take a look at iw output, and see if this god-awful expression still works
-	iw "$phyname" info 2>&1 | sed -e '/MHz/!d; /GI/d; /disabled/d; /radar detect /d; /Supported Channel Width/d; /STBC/d; /PPDU/d; /MCS/d; s/[:blank:]*\*[:blank:]*//g; s:[]()[]::g; s/\..*$//g' | awk ' { print "nextCh.push("$3"); nextChFreq["$3"] = \""$1"MHz\"; nextChPwr["$3"] = "$4";"   ; } ' >> "$out"
+	iw "$phyname" info 2>&1 | sed -e '/MHz/!d; /GI/d; /disabled/d; /radar detect /d; /Supported Channel Width/d; /STBC/d; /PPDU/d; /MCS/d; s/[:blank:]*\*[:blank:]*//g; s:[]()[]::g; s/\..*$//g' | grep  -v 'MHz -' | awk ' { print "nextCh.push("$3"); nextChFreq["$3"] = \""$1"MHz\"; nextChPwr["$3"] = "$4";"   ; } ' >> "$out"
 
 	echo "mac80211Channels[\"$chId\"] = nextCh ;"     >> "$out"
 	echo "mac80211ChFreqs[\"$chId\"]  = nextChFreq ;" >> "$out"
@@ -146,8 +146,10 @@ echo "var wifiDevA=\"\";" >> "$out_file"
 if [ -e /lib/wifi/broadcom.sh ] ; then
 	echo "var wirelessDriver=\"broadcom\";" >> "$out_file"
 	echo "var GwifiN = false;" >> "$out_file"
+	echo "var GwifiAX = false;" >> "$out_file"
 	echo "var AwifiN = false;" >> "$out_file"
 	echo "var AwifiAC = false;" >> "$out_file"
+	echo "var AwifiAX = false;" >> "$out_file"
 	echo "var dualBandWireless=false;" >> "$out_file"
 elif [ -e /lib/wifi/mac80211.sh ] && [ -e "/sys/class/ieee80211/phy0" ] ; then
 	echo 'var wirelessDriver="mac80211";' >> "$out_file"
--- a/package/gargoyle/files/www/basic.sh
+++ b/package/gargoyle/files/www/basic.sh
@@ -762,7 +762,6 @@ var isb43 = wirelessDriver == "mac80211" && (!GwifiN) ? true : false ;
 							<option value="ap+wds">AP+WDS</option>
 							<option value="sta"><%~ Clnt %></option>
 							<option value="ap+sta"><%~ Clnt %>+AP</option>
-							<option value="adhoc">Ad Hoc</option>
 							<option value="disabled"><%~ Disabled %></option>
 						</select>
 					</span>
--- a/package/gargoyle/files/www/basic.sh
+++ b/package/gargoyle/files/www/basic.sh
@@ -208,8 +208,9 @@ var isb43 = wirelessDriver == "mac80211" && (!GwifiN) ? true : false ;
 					<label class="col-xs-5" for="bridge_hwmode" id="bridge_hwmode_label"><%~ BrOpr %>:</label>
 					<span class="col-xs-7">
 						<select id="bridge_hwmode" class="form-control" onchange="setHwMode(this)">
-							<option value="11gn">B+G+N</option>
-							<option value="11g">B+G</option>
+							<option value="11gnax">G+N+AX</option>
+							<option value="11gn">G+N</option>
+							<option value="11g">G</option>
 							<option value="11anacax">A+N+AC+AX</option>
 							<option value="11anac">A+N+AC</option>
 							<option value="11an">A+N</option>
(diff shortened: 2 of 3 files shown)


package/gargoyle/files/usr/lib/gargoyle/cache_basic_vars.sh
```diff
@@ -95,7 +95,7 @@ print_mac80211_channels_for_wifi_dev()
 	# however, as far as I can tell there is no other way to get max txpower for each channel
 	# so... here it goes.
 	# If stuff gets FUBAR, take a look at iw output, and see if this god-awful expression still works
-	iw "$phyname" info 2>&1 | sed -e '/MHz/!d; /GI/d; /disabled/d; /radar detect /d; /Supported Channel Width/d; /STBC/d; /PPDU/d; /MCS/d; s/[:blank:]*\*[:blank:]*//g; s:[]()[]::g; s/\..*$//g' | awk ' { print "nextCh.push("$3"); nextChFreq["$3"] = \""$1"MHz\"; nextChPwr["$3"] = "$4";"   ; } ' >> "$out"
+	iw "$phyname" info 2>&1 | sed -e '/MHz/!d; /GI/d; /disabled/d; /radar detect /d; /Supported Channel Width/d; /STBC/d; /PPDU/d; /MCS/d; s/[:blank:]*\*[:blank:]*//g; s:[]()[]::g; s/\..*$//g' | grep  -v 'MHz -' | awk ' { print "nextCh.push("$3"); nextChFreq["$3"] = \""$1"MHz\"; nextChPwr["$3"] = "$4";"   ; } ' >> "$out"
```
lukasz1992, line 98: drop 802.11ah channels (shown as negative numbers)

package/gargoyle/files/usr/lib/gargoyle/cache_basic_vars.sh
```diff
@@ -146,8 +146,10 @@ echo "var wifiDevA=\"\";" >> "$out_file"
 if [ -e /lib/wifi/broadcom.sh ] ; then
 	echo "var wirelessDriver=\"broadcom\";" >> "$out_file"
 	echo "var GwifiN = false;" >> "$out_file"
+	echo "var GwifiAX = false;" >> "$out_file"
```
lukasz1992, line 149: did anyone noticed GwifiAx is undefined and breaks JS?

package/gargoyle/files/www/js/basic.js
```diff
@@ -104,7 +104,12 @@ function saveChanges()
 			{
 				uci.set("wireless", wifiDevG, "htmode",  getSelectedValue("wifi_channel_width") );						//always set htmode, even if it is "NONE"
 			}
-			txPowerSet("wifi_max_txpower", "wifi_txpower", wifiDevG)
+			txPowerSet("wifi_max_txpower", "wifi_txpower", wifiDevG);
+			if (getSelectedValue("wifi_hwmode_2ghz") != "11g")
+			{
+				uci.set("wireless", wifiDevG, "require_mode", "n");
+				uci.set("wireless", wifiDevG, "greenfield", "1");
```
lukasz1992, line 111: require_mode just rejects auth&assoc requests, greenfield drops the traffic as noise)

lantis1008:
Hey Lukas, thanks for the contribution. Really keen to check this out in detail and I'll get back to you.

lantis1008:
Thanks for the contribution @lukasz1992 . Unfortunately I can't accept the changes as is. Please find my feedback below. If you have any questions please ask.
5ca55ed - Enable WiFi 6 on 2.4GHz too, drop WiFi 2 settings
Minor changes required before accepting
- package/gargoyle/files/usr/lib/gargoyle/cache_basic_vars.sh
  - no issues
- package/gargoyle/files/www/basic.sh
  - L331,931,1014: Please leave channel 14 in so that if someone decides to enable B mode again it will be available
- package/gargoyle/files/www/js/basic.js
  - L2951-2952: Missing 11g/G mode here in modes/mnames respectively

1831d4b - Get rid of Ad-hoc mode
Accepted, no changes

38ab334 - Get rid of WEP encryption
Minor changes required before accepting
- package/gargoyle/files/usr/lib/gargoyle/cache_basic_vars.sh
  - Please keep the var wpad_wep line

e9c47f5 - Get rid of WPA version 1
Not accepted. Please remove this change.
I understand that WPA is not recommended, however it is still supported upstream at this stage and I'd like to stay close to that.

2aad650 - Allow newer encryptions for guest/bridge
Minor changes required before accepting
- package/gargoyle/files/www/basic.sh
  - L946: You can't have mixed encryption for connecting to other networks, only for broadcasting. It's either one or the other? It might work, but it doesn't make sense.

6dde1f5 - Drop support for WiFi 3
Not accepted. Please remove this change.
There are still uses for G and A modes. I would accept another option (under Connection -> Advanced) to set a minimum client capability (_require_mode_).
I won't accept _greenfield_. I feel this will create issues with peoples networks that they don't understand and can't debug. Maybe in the future.

faa8a4f - Reorder mode list; Channel 149 is fine for 160MHz width
Changes required before accepting
- package/gargoyle/files/www/basic.sh
  - L805-807: Please keep the old order
- package/gargoyle/files/www/js/basic.js
  - L3262: Is there ever an occasion when this will turn up as usable? I can accept it, but i'm wondering about the utility?

lukasz1992:
@lantis1008 
Thank you very much for detailed feedback, I will try to make changes in this week in this PR.

1.
- L331,931,1014: Please leave channel 14 in so that if someone decides to enable B mode again it will be available
- L805-807: Please keep the old order
A:
- I can revert the changes, I see anyway that this static content could be empty, as it is not being used - the list is being generated anyway (basing on cached vars to instert channels) and static content from .sh file is being replaced anyway. So channel 14 appears if radio0 supports it
I do not agree about channel 14. Keep in mind that with only user/web interface (and without access to SSH) you can not enable 802.11b, as OpenWrt disables it by default. Also channel 14 is allowed only in this scenario:
- you are in Japan
- you have 802.11b only mode (for WiFi 3 and next it is actually forbiden)
- you can hardly force OpenWrt to use 802.11b only mode; with this specific customization (https://forum.openwrt.org/t/how-to-set-legacy-802-11b-only-mode-since-21-02-2/124662/3)

2.
- L3262: Is there ever an occasion when this will turn up as usable? I can accept it, but i'm wondering about the utility?
A:
149 is a valid 160MHz channel. So do not limit channel 149 if cfg80211/mac80211 layer allows it.
Channel 149 (160MHz) is fine in the USA (only), basing on this: https://en.wikipedia.org/wiki/List_of_WLAN_channels

I will patch my PR to your review in all other points.

lantis1008:
Thanks for that. 
Sure, you can drop 14. Please fix the other ordering issue. 
I agree that the content is dynamically generated anyway and so maybe should be blank to save a few previous bytes. I might consider this in the future. 

Sure, 149 is fine 😃

lukasz1992:
@lantis1008 hi, please check again :) no work for Connection -> Advanced yet.

package/gargoyle/files/www/basic.sh
```diff
@@ -944,8 +944,12 @@ var isb43 = wirelessDriver == "mac80211" && (!GwifiN) ? true : false ;
 					<span class="col-xs-7">
 						<select class="form-control" id="wifi_encryption2" onchange="setWifiVisibility(); proofreadPass('wifi_pass2', this)">
 							<option value="none"><%~ None %></option>
+							<option value="sae-mixed">WPA3/WPA2 SAE/PSK</option>
```
lantis1008, line 947: Is this still right? My feedback was that mixed modes don't make sense for connecting to another network.

lantis1008 (commented):
Just 1 minor piece that may not have been addressed?
After that I need to do testing from a functionality point of view. Otherwise code looks good

package/gargoyle/files/www/basic.sh
lantis1008: Yes to 364.
Correct me if i'm mistaken, wifi_encryption2 is used only when we are joining an upstream wifi as a client right? So mixed would not make sense here either? Maybe i have this wrong.

lukasz1992: @lantis1008 yes, agree - gonna update PR again

lantis1008:
Thanks for taking on these changes and the feedback

PR #981, opened 2023-04-09
1.15 cleanup - default packages

aimacintyre:
Having already encountered examples of default packages not being included in images (e.g. swconfig in some ath79 images and kmod-leds-gpio in some mt76x8 images), adjust all target profile configs to ensure other required default packages are included in respective target images.

Also disable 2 ath79 devices for which we can't build factory images due to a much smaller image size limit than is applied to these devices' sysupgrade images (6144kB vs 7808kB).


lantis1008:
Did you do this by hand or did you write a script?
I might write a dev-util for checking this for the future.

aimacintyre:
I wrote a script to check the configs based on info copied from the various makefiles.  The info is structured in the script such that it could fairly easily be moved out to a JSON file.  I made the actual changes manually, as there weren't that many, but the script could easily be extended to apply the adjustments.  Happy to contribute the script if it would be useful longer term.

PR #982, opened 2023-04-12
additional devices, qosmon patch

aimacintyre:
Add devices:
- mt7621: Mercusys MR70X v1 and TP-Link Archer AX23 v1
- ath79: Belkin F9J1108 v2

Bump OpenWrt to 22.03.4 (required to be able to add mt7621 devices above).

Patch by Lantis1008 to improve behaviour of qosmon.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -34,7 +34,7 @@ set_version_variables()
 	# set precise commit in repo to use 
 	# you can set this to an alternate commit 
 	# or empty to checkout latest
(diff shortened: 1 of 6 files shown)


lantis1008:
@pbix courtesy heads up on a qosmon change due to freezing up of the loop under some circumstances on some devices. 
Understand you have moved on, so please treat this as FYI only unless you'd like to offer an opinion.

PR #983, opened 2023-04-14
update generic patches for OpenWrt 22.03.4

aimacintyre:
The following changes were required to get clean builds with the bump to OpenWrt 22.03.4:
- 011-iproute2_add-missing-libbpf-dependency.patch: delete - fixed upstream
- 110-gcom-getstrength.patch - refresh to account for upstream changes to target file


aimacintyre:
added patch to build.sh aborting build if any generic or target patches fail to apply.

PR #992, opened 2023-12-13
bump to OpenWrt 22.03.6

aimacintyre:
Due to a kernel backport of size_add(), applying the ntfs3 package patch 0009-Backport_size_add.patch results in a symbol redefinition; as the 2 implementations appear to be identical drop this patch.


lantis1008:
Much appreciated!

PR #1005, opened 2024-04-11
Gargoyle UI: add packet steering option to Connection/Advanced page

aimacintyre:
Add the ability to enable/disable Packet Steering as in OpenWrt's
Interfaces/Global Network Options page.

Packet steering is only useful for multi-cpu devices so don't show
the option if the device has only 1 cpu.

Polish is the only language plugin with translations for the Advanced options
page so update it (translations produced using Google Translate) for packet
steering.


lantis1008:
@obsy FYI some translations. Also if you hadn't noticed there is a branch for OpenWrt 23.05 based Gargoyle that is in a reasonably good place at the moment.

obsy:
I will update the translations in a few days.

obsy:
Is there any planned release date for the new version?

lantis1008:
Not yet. A few more things to do, then I might poke Eric on a release.

PR #1012, opened 2025-01-05
Switch to OpenWrt 23.05

lantis1008:
**Gargoyle 1.15.x is based on OpenWrt 23.05, which is a significant leap forward from 22.03 for 1.14.x.**
Configs should generally not be preserved between 1.14 (and earlier) and 1.15.x. Do so at your own risk.

A big thanks to pythonic (Github: aimacintyre) for their assistance and many contributions towards getting this into a position to make it ready for testing.

### Notable changes:
- Update from OpenWrt 22.03 -> 23.05
- Latest security updates
- ntfs3 has been switched to the in-kernel version
- Target "atheros" (ath25) has been removed. This has been marked as "source-only" upstream
- Switch from OpenSSL to MbedTLS for all default packages. This will save a heap of space on all devices. Some optional plugins still require it which may require external storage to fit
- Spectrum Analyser plugin minimal version has been dropped. The space savings were not worth the additional maintenance effort
- libmatrixssl package has been dropped
- OpenVPN (and EasyRSA) updated to support OpenSSL 3.x
- OpenVPN BF-CBC cipher has been removed (deprecated)
- New package MbedTLS Command Line Utility (mbedtls-clu). This is a brand new package designed to replicate (some of) the features of the OpenSSL Utility (openssl). It is smaller and relies on MbedTLS instead of OpenSSL so again, saves a bunch of space
- OpenVPN EasyRSA modified to work with mbedtls-clu
- Tor plugin dropped by default from all profiles. It is still available as a plugin.
- New DDNS IP Lookup provider ip2location.io
- New GeoIP Lookup provider ip2location.io
- DDNS updated to support a "test domain" where the configured update domain may not match the domain needed to detect the IP correctly (e.g. CloudFlare)
- Show additional information in DDNS GUI to make it easier to see which provider is being used
- Update OpenVPN and Wireguard to support the "test domain" variable from DDNS
- Add support for OpenVPN plugin to advertise additional subnets behind the server (not just the LAN subnet). This can be useful if you have cascaded LANs or want to allow access upstream as well
- New plugin DNS over HTTPS
- Add WAN port speed on the Status Overview page
- Add new subtarget Mediatek Filogic which includes some very popular new devices e.g. GL.iNet Flint 2 (MT6000)
- mvebu devices are back! These were previously disabled due to an upstream issue which has now been resolved
- Added UPnP IPv6 "Pinholes" to the GUI
- Added support for setting WAN VLAN to GUI
- Tor now supports IPv6 when operating in relay mode. It will block IPv6 connections when operating in other modes to prevent address leakage
- Bandwidth monitoring can now be disabled from the GUI

### Bugs squashed
- Quotas Others (Individual) not working
- Quotas bandwidth throttling using QoS not working in the upload direction
- Quotas bandwidth throttling not using QoS not working
- DDNS sometimes sending IPv6 updates to IPv4 providers (and vice versa)
- Some devices showing an empty temperature value on Status Overview
- Units (B, KB, MB, GB, TB etc) not selecting properly in some circumstances with language plugins other than English-EN
- Wireguard peer subnet routing not working in both directions
- Polish-PL plugin typo in Wireguard
- Added a check to DHCP hostnames to stop dnsmasq crashes
- Fixed a segfault with opkg (caused by gpkg)
- Fixed a bug causing webmon/weburl to miss a large proportion of matches
- Fixed gpkg not respecting --tmp-dir option
- Fixed a memory leak in webmon/weburl leading to devices crashing after a period of time

### Things to Note
- The "domain" setting has been switched from the default of ".lan" to ".home.arpa" in line with RFC 8375
- WiFi interfaces are no longer named "wlanX"
- Similarly a guest network is no longer "wlan0-1", it is now "wl0-sta0" and "wl0-ap0" for an Access Point configuration. This won't affect most users, but for anyone poking around be aware of this.
- [mbedtls-clu](https://github.com/lantis1008/mbedtls-clu) is a brand new utility which has not received wider testing and scrutiny. If you note any issues or differences between it and openssl, please raise an issue
- Bandwidth monitoring has been split into separate rule files which are loaded on restart. This allows the rules to be dynamically restarted by many different processes which solves a bug with QoS graphs sometimes not appearing

### New Devices
- Xiaomi Mi Router 4c
- DLink DIR-859-A3
- DLink DIR-869-A1
- TPLink WR841HP-v2/v3
- TPLink WR941HP-v1
- AVM Fritzbox 7520
- GL.iNet A1300
- Netgear SRR60
- Netgear SRS60
- Teltonika RUTX50
- Asrock G10
- Acer Predator W6 (Note: No 6GHz support in GUI)
- ASUS TUF AX4200
- ASUS TUF AX6000
- Cudy WR-3000-v1
- GL.iNet MT3000
- GL.iNet MT6000
- Netgear WAX220
- ASUS RT-AC57U-v1
- ASUS RT-AX54
- DLink DIR-3060-A1
- DLink DIR-853-A1
- Linksys e7350
- Linksys RE7000
- Netgear EX6150
- TPLink ER605-v2
- Zyxel WSM20
- Netgear WAX206
- Netgear WAX220
- Asus RT-AX59U
- BananaPiR3
- Xiaomi Redmi Router AX6000
- Ubiquiti Edgerouter-X-SFP
- Beeline Smartbox GIGA
- Several others I've forgotten to mention :)

### Known Issues
- If you have installed and enabled DNS over HTTPS plugin and then sysupgrade to an image without it (e.g. any image that you haven’t self compiled) you will have no DNS resolution until you manually fix your /etc/config/dhcp file

**Note:** OpenWrt 22.03 moved to nftables from iptables. Gargoyle still uses iptables to support the custom modules it needs, and so there is some package incompatibility with the base openwrt package repository. For normal operation this should not cause a problem. For anyone trying to install extra packages manually, your mileage may vary!


lantis1008:
I will merge this within the week. 
Open to allow comments and feedback.

lantis1008:
I am working in the background on switching to nftables. Once this is done I will move to 24.10 so that we don’t have to keep carrying a significant divergence from upstream. 
if we don’t do it this way, nftables AND apk will need to be done at the same time.

nftables is a significant undertaking and the custom kernel modules must be rewritten. 

The point of merging to master is so I can start a new branch for 24.10 as well. 
the 23.05 builds have been available for a long time on our forums.

obsy:
I was more thinking about the fact that it was 24.10 still with iptables (opkg is standard in 24.10).

lantis1008:
I understand. But if I don’t do at least one of those before doing 24.10, I have to do both for 25.xx.

lantis1008:
The effort of staying on iptables is frustrating. For every update I need to check if upstream did something nftables only and revert it or rewrite it. There are bugs in fw3 that won’t be solved unless move to fw4. They are written in ucode so back porting fixes is too hard. 
for every package I need to be careful. Some authors have both iptables and nftables, most are lazy of course.

I am worried that if I don’t do it now it will never happen and eventually it will not be optional.

PR #1017, opened 2025-04-14
fon-flash: fix aarch64 where char can be unsigned

cho-m:
In Homebrew, we are currently building binaries for aarch64 (armv8-a) on GitHub's `ubuntu-22.04-arm` runner using the system GCC 11. The resulting `fon-flash` binary ended up never exited.

This seems like common issue of porting to ARM where `char` can be unsigned - https://developer.arm.com/documentation/den0013/d/Porting/Miscellaneous-C-porting-issues/unsigned-char-and-signed-char

I think the type can just be made into `int` like official documentation for `getopt()`

### DIFF
--- a/fon-flash/fon-flash.cpp
+++ b/fon-flash/fon-flash.cpp
@@ -1231,7 +1231,7 @@ int main(int argc, char* argv[])
 	int num_files_expected = 0;
 	int num_files_found = 0;
 	
-	char c;	
+	int c;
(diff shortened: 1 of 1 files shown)


lantis1008:
Thanks for the contribution. 
Have you tested the binary functionally and it still works?
I don't have any way of verifying this on my end.

lantis1008:
On looking into this further, it is a pretty innocuous change, so I have merged it.
Thanks again for your contribution.

